PR #2, opened 2022-01-16
PB Updates

brianhealey:
This draft PR contains  updates to the protocol buffer schema exposed by the tesla powerwall+ (and hopefully others) api's.

These changes  are breaking protocol buffer changes, and will require some code to that uses it to be updated.


examples/vitals/pull_vitals.py
```diff
@@ -53,13 +53,80 @@
...
-    print("     - %s" % pw.devices[x].device[0].device.din.value)
+    device = pw.devices[x].device.device
+    parent = str(device.componentParentDin.value)
+    vitals = pw.devices[x].vitals
+    alerts = pw.devices[x].alerts
+
+    name = str(device.din.value)
+    print("Device %d: %s " % (x, name))
+
+    # e.STSTSM = "STSTSM",
```
brianhealey, line 64: I don't have a direct translation for these types. If they are known I would love to add friendly values to represent them to the callers.

examples/vitals/pull_vitals.py
```diff
@@ -53,13 +53,80 @@
...
+            print("   - Alerts:")
+
+        print("       - ALERT_%i = %s" % (a, alerts[a]) )
+        a += 1
+
+    print("   - Vitals:")
+
     for y in pw.devices[x].vitals:
         vital_name = str(y.name)
+        if (y.HasField('intValue')):
```
brianhealey, line 127: I don't actually receive any that have this, but it is in the spec, so it might as well be handled.

examples/vitals/tesla.proto
```diff
@@ -9,8 +9,277 @@ syntax = "proto3";
...
 package teslapower;
 
+import "google/protobuf/any.proto";
 import "google/protobuf/timestamp.proto";
 
+message ExternalAuth {
+    ExternalAuthType type = 1;
+}
+
+enum ExternalAuthType {
```
brianhealey, line 19: Most of these aren't directly related to the /api/devices/vitals endpoint. I documented them here in case someone else finds them useful

examples/vitals/tesla.proto
```diff
@@ -43,17 +324,19 @@ message TeslaHardwareId {
 }
 
 message TeslaEnergyEcuAttributes {
-    optional int32 ecuType = 1;
+    int32 ecuType = 1;
     optional TeslaHardwareId hardwareId = 2;
+    optional PVInverterAttributes pvInverterAttributes = 3;
```
brianhealey, line 329: I only ever see `ecuType` populated in my install, but the spec had these also in case others do return them.

examples/vitals/tesla.proto
```diff
@@ -80,19 +363,162 @@ message Device {
     optional google.protobuf.Timestamp firstCommunicationTime = 8;
     optional google.protobuf.Timestamp lastCommunicationTime = 9;
     optional ConnectionParameters connectionParameters = 10;
-    repeated DeviceAttributes deviceAttributes = 11;
+    optional DeviceAttributes deviceAttributes = 11;
```
brianhealey, line 366: Breaking change. Single value only.

examples/vitals/tesla.proto
```diff
@@ -80,19 +363,162 @@ message Device {
...
 }
 
 message SiteControllerConnectedDevice {
     optional Device device = 1;
 }
 
 message SiteControllerConnectedDeviceWithVitals {
-    repeated SiteControllerConnectedDevice device = 1;
+    SiteControllerConnectedDevice device = 1;
```
brianhealey, line 374: Breaking change. I had this wrong in my initial attempt.

examples/vitals/pull_vitals.py
jasonacox: Same here, but I agree.  Would love to hear from others if you are getting these data points.

examples/vitals/tesla.proto
jasonacox: Nice! Makes sense and this cleans up the referencing.

examples/vitals/tesla.proto
jasonacox: Makes sense.  For all of these updates, for pypowerwall, I think we can accomodate without major API impact to the pypowerwall functions (as minor rev) but since it could impact vitals() output, I'll increment as a major version rev. to signal caution.

jasonacox (approved):
Missing: Will need to fix __init__.py - need to remove the [0] indexing which is no longer needed due to the revised PB schema:
```python
            parent = str(pb.devices[x].device[0].device.componentParentDin.value)
            name = str(pb.devices[x].device[0].device.din.value)
```
to
```python
            parent = str(pb.devices[x].device.device.componentParentDin.value)
            name = str(pb.devices[x].device.device.din.value)
```

jasonacox:
Thank you @brianhealey !

slcasner:
It appears that connectionParameters is not correct.  The only instance that is populated in my vitals output is for the NEURIO where field called ipAddress has the value PWRview-11178.  The following is an excerpt from the DebugString() output:

devices {
  device {
    device {
      din {
        value: "NEURIO--VAH4602AB0636"
      }
      serialNumber {
        value: "VAH4602AB0636"
      }
      manufacturer {
        value: "NEURIO"
      }
      componentParentDin {
        value: "STSTSM--1232100-00-G--TG121253002Z69"
      }
      firmwareVersion {
        value: "1.6.1-Tesla"
      }
      lastCommunicationTime {
        seconds: 1638768324
      }
      connectionParameters {
        ipAddress {
          value: "PWRview-11178"
        }
      }
      deviceAttributes {
        meterAttributes {
          meterLocation: 5
        }
      }
    }
  }
  vitals {
    name: "NEURIO_CT0_Location"
    stringValue: "solarRGM"
  }
}

jasonacox:
Hi @slcasner - if you run this below, it looks like this is accurate.  The ipAddress for the Neurio seems to be "PWRview-" with ShortID suffix.  Using the /api/meters API call (not the protobuf /vitals data) you will see the same thing. 

```python
import pypowerwall
import json

pw = pypowerwall.Powerwall(host,password,email,timezone)
pw.poll('/api/meters')
```
Snip of Neurio section:
```json
{
    "serial": "VAHXXXXXXXXXX",
    "short_id": "XXXXX",
    "type": "neurio_w2_tcp",
    "connected": true,
    "cts": [
        {
            "type": "solarRGM",
            "valid": [
                true,
                false,
                false,
                false
            ],
            "inverted": [
                false,
                false,
                false,
                false
            ],
            "real_power_scale_factor": 2
        }
    ],
    "ip_address": "PWRview-XXXXX",
    "mac": "xx-xx-xx-xx-xx-xx"
}
```
I suspect it is a local lookup name on the Tesla gateway.

slcasner:
OK, I guess the element names ipAddress and ip_address are just inappropriately chosen, then.  To my mind, an IP address (for IPv4) should be a dotted quad numeric value.

[Sorry for the delayed reply, I was traveling.]

PR #3, opened 2022-01-21
Add backup switch and alert

dailow:


jasonacox:
Great addition... Thanks @dailow !

PR #8, opened 2022-03-05
Updates to __init__ and poll()

### LINKED ISSUE #5: Modification to poll() defaults
Two things related to poll()

1. When creating the Powerwall object, it would be good to pass in a cache_expiration parameter that can adjust the cache expiration
2.  It would be good to be able to override the cache with a Fetch=True parameter to poll()

If I make a pull request and make these changes, would you be interested in accepting them?

wcwong:
- Closes #5
 - Added pwcacheexpire to __init__ to allow adjustment of the cache time
 - Added timeout to __init__ to allow adjustment of the http timeout
 - added force to pol() to bypass the cache regardless of the timeout
 - moved the check for the vitals to after the cache check (minor optimization)

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,10 +99,12 @@ and call function to poll data.  Here is an example:
  set_debug(True, color=True)
 
  Classes
-    Powerwall(host, password, email, timezone)
+    Powerwall(host, password, email, timezone, pwcacheexpire, timeout)
 
  Functions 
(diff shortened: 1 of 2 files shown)


wcwong:
Let me know if I got this right...

jasonacox:
Thank you @wcwong !!  Great job!

PR #9, opened 2022-03-05
Implement call to /api/system_status/grid_status

### LINKED ISSUE #7: Add grid_status function
Provide a call to /api/system_status/grid_status 

from https://github.com/vloschiavo/powerwall2 the return values would be: 
```
{"grid_status":"SystemGridConnected","grid_services_active":false}
{"grid_status":"SystemGridConnected"} = grid is up
{"grid_status":"SystemIslandedActive"} = grid is down
{"grid_status":"SystemTransitionToGrid"} = grid is restored but not yet in sync.
```

if `verbose=True` then return the full json which would include `grid_services_active`

wcwong:
- Closes #7
 - add function grid_status() that matches the API call

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,27 +102,30 @@ and call function to poll data.  Here is an example:
     Powerwall(host, password, email, timezone, pwcacheexpire, timeout)
 
  Functions 
-    poll(api, json, force)  # Return data from Powerwall API URI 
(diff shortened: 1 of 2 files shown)


jasonacox:
Looks good.  Thanks @wcwong !

PR #14, opened 2022-05-20
Pass through Powerwall Web interface and customize for iframe displays

danisla:
Hello,

This PR adds support to the proxy feature of pypowerwall to proxy access to the Powerwall web interface and customize it's appearance. 

This lets you run the proxy on a local network and create iframe widgets for displays like Dakboard.

image

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.7-alpine
 WORKDIR /app
-RUN pip3 install pypowerwall==0.4.0
+RUN pip3 install pypowerwall==0.4.0 bs4
 COPY . .
 CMD ["python3", "server.py"]
 EXPOSE 8675 
--- a/proxy/server.py
+++ b/proxy/server.py
@@ -24,6 +24,9 @@ import sys
 import resource
 import requests
 
+from transform import get_static, inject_js
+web_root = os.path.join(os.path.dirname(__file__), "web")
+
 PORT = 8675
 BUILD = "t11"
 
@@ -212,6 +215,14 @@ class handler(BaseHTTPRequestHandler):
             self.send_header("Set-Cookie", "AuthCookie={};path=/;SameSite=None;Secure;".format(pw.auth['AuthCookie']))
             self.send_header("Set-Cookie", "UserRecord={};path=/;SameSite=None;Secure;".format(pw.auth['UserRecord']))
 
+            # Serve static assets from web root first, if found.
+            fcontent, ftype = get_static(web_root, self.path)
+            if fcontent:
+                self.send_header('Content-type','{}'.format(ftype))
+                self.end_headers()
+                self.wfile.write(fcontent)
+                return
+
             # Proxy request to Powerwall web server and cache.
             cache_item = web_cache.get(self.path, None)
             if not cache_item:
@@ -230,6 +241,11 @@ class handler(BaseHTTPRequestHandler):
                 web_cache[self.path] = (fcontent, ftype)
             else:
                 fcontent, ftype = cache_item
+
+            # Inject transformations
+            if self.path.split('?')[0] == "/":
+                if os.path.exists(os.path.join(web_root, "customize.js")):
+                    fcontent = bytes(inject_js(fcontent, "customize.js"), 'utf-8')
 
             self.send_header('Content-type','{}'.format(ftype))
             self.end_headers()
--- a/proxy/server.py
+++ b/proxy/server.py
@@ -22,6 +22,7 @@ import json
 import time
 import sys
 import resource
+import requests
 
 PORT = 8675
 BUILD = "t11"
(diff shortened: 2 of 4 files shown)


jasonacox:
Thank you, @danisla!  This is a great addition!  A few things in my testing:

- I had to update the content-type for the json payloads for Chrome and Brave browsers.  I noticed you had it comment out and I changed the logic to dynamically set content-type.  I need to verify that this doesn't negatively impact telegraf or other polling tools (which it shouldn't)
- I am not able to get this to work in Safari browser.  I haven't had time to explore why, but it is going to the login screen which of course the proxy is not set to support (POST).  Let me know if you have any suggestions.
- I am going to soak test the proxy.  Many people have this running 24x7 so I need to ensure it doesn't have a memory/resource leak with the addition of the great new features.

Great job on this!  Thanks for the PR.

jasonacox:
Update:   Another issue I found is that the pass through would not work if you used a hostname or IP address.  Using localhost worked on Chrome and Brave browsers.  I was able to get it to work for all hostnames and on Safari.  I had to remove the `send_header()` attributes "path=/;SameSite=None;Secure;" From what I can tell, that is only proper if you use HTTPS but since the proxy is HTTP, they were causing the browser to ignore the auth cookies.  I added the ability to make the proxy serve HTTPS via an environmental setting.  I need to run more tests and build an iframe demo to ensure all cases and all browsers work (http and https).  I suspect there are still issues with iframes.

Can you see if the changes still work with your Dakboard?

danisla:
@jasonacox thanks for the feedback!
I run the pypowerwall proxy in http mode behind an https proxy and I missed the localhost and IP cases, thanks for catching those.

For ideal iframe support, the proxy would need to run in https mode with a valid cert and have the `SameSite=None;` and `Secure;` cookie suffixes.

I tested this with `PW_HTTPS=yes` on my Dakboard and it wouldn't load with the self-signed certificate, there isn't really a way to bypass the cert exception on the iframe without modifying the root certs. 

Maybe if the pypowerwall proxy had an option to set the https mode cookiesuffix, while running in http mode, that would fix it for my case where my ssl termination is done upstream. I'm open to other solutions too. 

Another thing I noticed is that the cookie should be set with the `path=/;` prefix in all cases, otherwise you get multiple cookies set for the /api paths too:
```python
if(https_mode == "yes"):
    cookiesuffix = "path=/;SameSite=None;Secure;"
    httptype = "HTTPS"
else:
    cookiesuffix = "path=/;"
    httptype = "HTTP"
```

Also FYI, the web app has an indexedDB requirement which means the web app doesn't work in some browsers or in incognito  where indexedDB is blocked. (I was running my dakboard on an RPI with Chromium in incogntio and the web app iframe wouldn't load, but disabling incognito fixed it)

jasonacox:
Thanks @danisla!  I think we could have a `PW_HTTPS=http` mode that adds the SameSite suffix without adding another env:

```python
if https_mode == "yes":
    # run https mode with self-signed cert
    cookiesuffix = "path=/;SameSite=None;Secure;"
    httptype = "HTTPS"
elif https_mode == "http":
    # run http mode but simulate https for proxy behind https proxy
    cookiesuffix = "path=/;SameSite=None;Secure;"
    httptype = "HTTP"
else:
    # run in http mode
    cookiesuffix = "path=/;"
    httptype = "HTTP"
```

Is there a way to detect if indexedDB is blocked?  If so, we could have it send a message to the user to disable incognito.

jasonacox:
Hi @danisla  - I made some updates:

* Proxy can now run in HTTP mode but will send SameSite data (simulating https).  Hopefully this works for your use case where pypowerwall is behind an https proxy. You would add `-e PW_HTTPS='http'` to your `docker run` or equivalent to set the environmental variable. 
* Fixed the missing 'path=/' suffix.  Thank you!
* Added pass through feature notes to the [README.md](https://github.com/jasonacox/pypowerwall/tree/main/proxy#power-flow-animation---passthrough)
* Added a iFrame example to the proxy web root - [example.html](https://github.com/jasonacox/pypowerwall/blob/main/proxy/web/example.html) to make it easier for others to see how to embed in their projects.
* Fixed several bugs (!!!) that I discovered doing all of the above. 

I'm now soak testing the latest build to ensure no memory leaks or other issues.  

Again, THANK YOU for all your help and the great contribution!

danisla:
Awesome, glad to be of help.

Thank you for creating and maintaining this project!

jasonacox:
One update in case you pull, I changed the js injection to use clear.js which uses 'transparent' background to make it easier to place on multiple projects.  I made this a configurable option via `PW_STYLE="clear"` environmental variable that pulls in the corresponding js file.  Options are: `clear` (default), `black`, `white`, `grafana` gray, and `dakboard` black (your original). It is not an elegant solution but does work.

FYI - I also released an updated Grafana dashboard thanks to your help: https://github.com/jasonacox/Powerwall-Dashboard/discussions/26

Thanks again, Dan!

PR #15, opened 2022-05-23
remove scrollbars from web view

danisla:
Sometimes when embedding in an iframe, the vertical scrollbars are visible.

This patch hides the scrollbars for all styles.

Before:
image

After:
image


jasonacox:
Thanks Dan!

PR #16, opened 2022-07-03
Support specifying a bind address

zi0r:
The current code always binds to 0.0.0.0.  This adds support to specify binding (for example) to 127.0.0.1 to ensure the proxy is only accessible via localhost.

Example:
`export PW_BIND_ADDRESS=127.0.0.1;python ./server.py`


jasonacox:
Thanks @zi0r !  Nice addition.

PR #17, opened 2022-07-03
Add shebang for direct execution

zi0r:
Add shebang to allow us to:
`chmod +x server.py;./server.py`


jasonacox:
Thanks @zi0r !  It needed this for sure.

PR #18, opened 2022-07-03
Add FreeBSD-specific installation instructions

zi0r:
I've created a port for pypowerwall to make installation easier for FreeBSD users.  Add instructions on how to use it.


jasonacox:
Thanks Ryan!

PR #19, opened 2022-07-21
Add grid_status responses for syncing to off-grid

mcbirse:
Added `SystemTransitionToIsland` and `SystemIslandedReady` to map, as these can be returned when transitioning to off-grid (more details in issue #7 )


jasonacox:
Thanks for the fix @mcbirse !  Great catch.

Issue #20, opened 2022-07-25
Implement HTTP persistent connections for API requests


mcbirse:
I've noticed API requests to the Powerwall gateway are not using HTTP persistent connections (aka keep-alive or connection re-use), which results in every request opening a new connection to the gateway and closing again.

There can be a lot of requests being sent to the gateway, even more so if the Power Flow is being viewed via the proxy as with the default Grafana dashboard of the Powerwall-Dashboard (~7 additional requests every few seconds).

I assume the constant open/close of HTTP connections might cause more load on the Powerwall, not to mention the continual requirement for new TCP sockets and such. My Powerwall has been a bit flaky sometimes where it stopped responding for a while, which could be attributed to overloading it perhaps?

So, to potentially reduce the load and make pypowerwall more efficient with API requests to the gateway, I was thinking it would be better to utilise HTTP connection re-use and pooling.

This appears to be simple enough by using a [Session object](https://requests.readthedocs.io/en/latest/user/advanced/) of the Requests module. I will submit a PR with my code changes for this.

Using the Session object results in HTTP connections to the gateway remaining established via keep-alive and being re-used with each API request. There could be multiple connections to the gateway (pooling), but this will be limited by the max pool size (10 by default, which seemed fine in my testing).

I ended up doing a fair amount of before/after testing of this change, as I wanted to quantify and validate that it was in fact beneficial?

I also did resilience testing with Powerwall-Dashboard and the pypowerwall proxy, to ensure there were no adverse affects. i.e. Testing what happens if an established HTTP connection was purposefully broken, invalidating the auth cookie to see if a re-login would occur, disrupting the network connection to the gateway for a while, etc. All my tests passed, with the connection to the gateway re-establishing okay without any user intervention or having to restart the pypowerwall proxy. :thumbsup:

Anyway, I thought I would include findings from my before/after testing for anyone interested...

@jasonacox - let me know what you think? This appears to be a very worthwhile change I believe. I think I just missed the v0.5.1 release as I was typing this out when released! v0.6.0 maybe? 😊 

And sorry for such a long post! But hopefully the below testing might also be informative to others (as it was for me).

---

Firstly, I wanted to see what happens regarding HTTP connections direct to the gateway using a standard web browser. As my Powerwall is in a DMZ on my network, I used my firewall to view the TCP connections.

[...]

PR #21, opened 2022-07-27
Use HTTP persistent connections for API requests

mcbirse:
Implemented the use of HTTP persistent connections for API requests by default (refer issue #20)

- Requests to Gateway will now re-use persistent http connections which reduces load and increases response time
- Default connection `poolmaxsize=10` to align with Session object defaults. Note: pool use applies to multi-threaded use of pyPowerwall only, e.g. as with the pyPowerwall Proxy Server
- Added env `PW_POOL_MAXSIZE` to proxy server to allow this to be controlled (persistent connections disabled if set to zero)
- Added env `PW_TIMEOUT` to proxy server to also allow timeout on requests to be adjusted
- For release prep, pyPowerwall version changed to v0.6.0 and proxy version to t17 (release notes still required)


jasonacox:
Thank you @mcbirse !  This is a great improvement with elegant approach and code.  Happy to merge!

PR #30, opened 2023-03-14
Additional alert message

DerickJohnson:
Added VFCheck_RoCoF alert from PVAC


jasonacox:
Thanks @DerickJohnson !

I suspect PVAC_a035_VFCheck_RoCoF has something to do with "rate of change of frequency" (RoCoF) df/dt

PR #32, opened 2023-03-21
Fix Cache-Control no-cache header and allow for setting max-age, fixes #31

### LINKED ISSUE #31: Proxy cache-control not set correctly
Setting Cache-Control in server.py [here](https://github.com/jasonacox/pypowerwall/blob/4dc18b0ab8c5b58dc1d84ca95e9083fb3249e7aa/proxy/server.py#L292) is not done correctly.  This line of code
```python
self.send_header("Set-Cookie", "Cache-Control: no-cache")
```
is incorrect as cache control is not a cookie, it is it's own header field.  The correct code is:
```python
self.send_header("Cache-Control", "no-cache, no-store")
```
Note that I added `no-store`.   I did this because when analyzing what the Testa Powerwall returns when I connect the browser directly to the Powerwall, I see both no-cache and no-store.  I believe this is to support older browsers.

Okay, so, by default pyPowerwall should set `no-cache, no-store` by fixing this bug.

But not caching has severe performance implications when reloading a page with embedded power flow animation graphics.  We should provide an option that lets a user indicate that they want to permit a browser to cache resources.  I propose a setting.
`PW_BROWSER_CACHE=<seconds>`
and if this is set to anything non-zero then
```python
browser_cache = int(os.getenv("PW_BROWSER_CACHE", "0"))
if browser_cache > 0:
    self.send_header("Cache-Control", "max-age={}".format(browser_cache))
else:
    self.send_header("Cache-Control", "no-cache, no-store")
```
This dramatically improves page load (after the first load) until whatever number of seconds expires and then the browser will fetch the resources again (on  page reload).  A user can always force the reload by holding down the shift key.

image

Notice the resources that are now retrieved from memory.

Now, we could further refine this.  Rather than setting the header before the data is retrieved from Powerwall, wait until after so that the content-type is known.  If content-type is `text/css`, `application/javascript` or `image/png` then use `max-age` else do not permit caching.  Specifically `application/json` should never be permitted to cache. This is what appears to be happening by default, but I don't know if that is coincidence or not...  It's not really coincidence (as coded right now) because that data is coming from `<ipaddr>:8675/api/xxxx` that are in the `ALLOWLIST` so it is not going through the proxy code.

So, it might be safer to move the header code down a few lines where we already have...
`ftype = r.headers['content-type']`
and then add
```python
browser_cache = int(os.getenv("PW_BROWSER_CACHE", "0"))
# Allow browser caching, if user permits, only for CSS, JavaScript and PNG images...
if browser_cache > 0 and (ftype == 'text/css' or ftype == 'application/javascript' or ftype == 'image/png'):
    self.send_header("Cache-Control", "max-age={}".format(browser_cache))
else:
    self.send_header("Cache-Control", "no-cache, no-store")
```
And finally, I would move the first line up to where all the other environment variables are retrieved.

David

dkerr64:
Fix for the issue described in #31

### DIFF
--- a/proxy/README.md
+++ b/proxy/README.md
@@ -91,6 +91,16 @@ The Proxy will pass authenticated calls through to the Powerwall Web Interface a
 
 This is available by directly accessing the proxy endpoint, https://localhost:8675 (replace localhost with the address of host running pyPowerwall Proxy). You can embed this animation within an iFrame. See [web/example.html](web/example.html).
 
+## Browser Cache Control
+
+By default resources sent for the power flow animation passthrough are not cached by the browser.  This includes fairly large CSS, JavaScript and Image PNG files which are downloaded every time the browser reloads the animation.  Performance can be improved by directing the web browser to cache these resources locally - only reloading if the data in the cache is old, a period known as `max-age`.
+
+You can control this with an optional environment variable `PW_BROWSER_CACHE` which takes a value in seconds. For example,
+
+* PW_BROWSER_CACHE=86400 - set `max-age` to 24 hours.
+
+If `PW_BROWSER_CACHE` is not set, or set to zero, then no caching takes place.  If you need to force a reload of the browser cache before `max-age` then most browsers will do this if you hold down the `shift` key while reloading the page.
+
 ## HTTPS Support (Experimental)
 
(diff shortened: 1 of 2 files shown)


jasonacox:
Brilliant!  Excellent addition @dkerr64 . Thank you! I'll run tests and push the container as `v0.6.1 Build t25`.

PR #36, opened 2023-04-15
Create tessolarcharge.py

venturanc:
1st version: Add script to allow charging a Tesla car with excess solar, using PyPowerwall local docker container to get realtime production and consumption data from Tesla Powerwall, and change car charging rate via Tesla API with TeslaPy.

ToDo: add environmental variables, possible add web API endpoints to turn off and on solar charging so that script can be left running indefinitely when car plugged in and do not want to charge.

### DIFF
--- a/tools/tessolarcharge.py
+++ b/tools/tessolarcharge.py
@@ -1,0 +1,272 @@
+"""
+ Adapted by Nate Carroll 03/2023 from TesSense w/ SenseLink  -Randy Spencer 2023 Version 9.7
+ Python charge monitoring utility for those who have a Tesla Powerwall.
+ Uses jasonacox/pypowerwall docker container to get realtime stats from the local gateway for Production 
+ and Grid Utilization of electricity to control
+ your main Tesla's AC charging amps to charge only with excess production.
+ Simply plug in your car, update your info below, and type> python3 tessolarcharge.py
+ 
+ Added: checking of location of Tesla to be sure it's charging at home
+ Added: tracks cabin temp and local chargers, vents the car if it gets too hot 
+ - removed due to Tesla removal of Vent option from API
+ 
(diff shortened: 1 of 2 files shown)


tools/tessolarcharge.py
```diff
@@ -0,0 +1,272 @@
...
+ 
+ Added: checking of location of Tesla to be sure it's charging at home
+ Added: tracks cabin temp and local chargers, vents the car if it gets too hot 
+ - removed due to Tesla removal of Vent option from API
+ 
+ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND ON INFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"""
+
+username = 'elon@tesla.com' # Fill in Tesla login email address/account
```
jasonacox, line 17: I'm sure Elon would appreciate if we just used the ICANN approved example address "yourname@example.com" instead :)

jasonacox:
This is great @venturanc !  Thank you!  We should create some documentation (I can help if you want) by adding a section to:

* https://github.com/jasonacox/pypowerwall/blob/main/tools/README.md

tools/tessolarcharge.py
venturanc: Ha ha, maybe if everyone charged his car with their solar he wouldn't need all of that Twitter verified money? But good idea, I'll change and submit another PR. :-)

venturanc:
Good idea, I'll see if I can add some details to the readme and you can fix anything I leave out? I'll copy the format of the other tools.

tools/tessolarcharge.py
jasonacox: Thanks!  You won't need to submit another PR.  Just commit and push your changes to the branch you used for this PR (`venturanc:main`).

jasonacox:
Thanks @venturanc ! This is a great addition. 🙏

venturanc:
I'm dumb and left out the requests import when I was copying everything from my test environment so the current version will give an error message if it's run during solar production hours (this is why I shouldn't develop in a Jupyter notebook).  Setting this up from scratch on a new environment and will test when the sun is shining (hopefully tomorrow) and I'll make sure I have it working before submitting an updated PR.  Sorry about that.

PR #37, opened 2023-04-20
Update tessolarcharge.py

venturanc:
fix errors in previous commit that stopped script from functioning correctly: 
-import requests (was causing error when trying to get solar/consumption data from pypowerwall), 
-fix UpdateSense() (error was unable to get car charging voltage), 
-fix error that would cause "stop charging" command to send again a second time an hour after night-time stop charging, 
-run main only if executing script.

### DIFF
--- a/tools/tessolarcharge.py
+++ b/tools/tessolarcharge.py
@@ -129,7 +129,7 @@ def UpdateSense() :  # Update Powerwall and charger voltage
         print(e)
         printmsg(RedTxt + "Powerwall data timeout or cannot get charger voltage" + NormTxt)
         power_diff = 0
-		volts = 240 #don't change anything in case next call recovers
+        volts = 240 #don't change anything in case next call recovers
         return(True)
         
 def Vent(car, command) :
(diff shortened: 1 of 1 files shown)


tools/tessolarcharge.py
```diff
@@ -125,13 +125,12 @@ def UpdateSense() :  # Update Powerwall and charger voltage
     try :
         power_diff = int(update_powerwall()['site']['instant_power'])*-1
         volts = int(vehicles[0].get_vehicle_data()['charge_state']['charger_voltage'])
-    except :
+    except Exception as e:
+        print(e)
         printmsg(RedTxt + "Powerwall data timeout or cannot get charger voltage" + NormTxt)
         power_diff = 0
+		volts = 240 #don't change anything in case next call recovers
```
jasonacox, line 132: This intent seems wrong.

tools/tessolarcharge.py
```diff
@@ -150,6 +149,7 @@ async def TesSolarCharge() :
             print('Use browser to login. Page Not Found will be shown at success.')
             print('Open this URL: ' + tesla.authorization_url())
             tesla.fetch_token(authorization_response=input('Enter URL after authentication: '))
+		global vehicles
```
jasonacox, line 152: Check this indent as well.  Seems like this would throw and exception.  I recommend running `pylint` to ensure there isn't anything off.

venturanc: yes, thanks for catching and your patience! Adding pylint to my workflow!

jasonacox:
Thanks @venturanc !

PR #38, opened 2023-04-27
Fix tessolarcharge.py script

venturanc:
Stability fix- don't stop charging a vehicle that is not charging. Tested and runs for days across multiple plug/unplug sessions, nights without charging, etc.


jasonacox:
Merged. Thanks @venturanc !

PR #40, opened 2023-05-18
Fix typos and spelling errors

mcbirse:
Just a few fixes to correct some typos and spelling errors, as well as STSTSM firmware actually shows "localbuild" for firmware 23.4.2-1 fe55682a (mistyped I would guess probably due to finger memory typing localhost a million times! 😄 )


jasonacox:
Thanks, @mcbirse ! 🙏

Issue #42, opened 2023-07-06
Alert Code feedback


ShawnKimble:
Adding some feedback on alert codes, and my understanding of what some of them mean:

BackfeedLimited - I believe this controls backfeed before PTO, under the "Permission to Operate" option under the settings menu.  Prior to PTO the backfeed rate is limited, as the system needs to produce over the load, but by the minimum possible.  The system seems to regulate this by switching MCI's open/closed, to minimize the overage that must be backfed, as the power has to go somewhere, and at this point the batteries are 100%.  This seems to toggle that setting per inverter.  I was told by Tesla that one of my two powerwalls were still stuck in this mode, post PTO, because of some software glitch that doesn't always toggle all devices to on, when selecting Yes to operate from the app.

RealPowerAvailableLimited  - Seems related to the above

POD_w031_SW_Brick_OV
POD_w044_SW_Brick_UV_Warning
POD_w045_SW_Brick_OV_Warning
I believe all the Brick warnings are related to preventing the condition where the powerwall doesn't have the minimum amount of power it needs to turn back on.  When this happens, a third party charger is needed to get the powerwall back to it's minimum operating battery requirement to turn back on, or it's "bricked".  Solar cannot return it to this state, because it needs power to make power.

jasonacox:
This is great, @ShawnKimble !!  

I made some updates based on your notes in a pull request here: https://github.com/jasonacox/pypowerwall/pull/43 - feel free to review and comment before I merge.

jasonacox-sam:
## Closing — Documentation Added ✅

**Issue summary:** Community-contributed documentation of Tesla Powerwall alert codes (BackfeedLimited, RealPowerAvailableLimited, POD brick warnings, POD_w110_SW_EOC, etc.) and their meanings.

**What happened:** @ShawnKimble provided helpful explanations of several alert codes, including how BackfeedLimited controls pre-PTO backfeed regulation via MCI toggling, and how POD brick warnings relate to minimum battery thresholds. Jason incorporated this feedback into [PR #43](https://github.com/jasonacox/pypowerwall/pull/43), updating the alert code documentation in the codebase.

This was an informational thread and the documentation improvements have been merged.

**Still have questions?** Feel free to reopen or start a new discussion. Any updates that help the community are always welcome.

---
*— Sam 🌟 (pypowerwall agent)*

PR #43, opened 2023-07-07
Add alert definitions per #42

jasonacox:


ShawnKimble:
PVS_a059_MciOpen - "Mid-Circuit Interrupter" is open, this happens when there is not enough solar power to turn on the string, or the emergency shut down button is pressed.  These are safety devices on the strings to turn them on and off.
PVS_a060_MciClose  - "Mid-Circuit Interrupter" is closed, this is normal operation.  An AC signal is sent from the inverter up the DC string triggering the MCI relay to close, allowing for DC solar production to start.

jasonacox:
Merging this into the README. Thanks for your discovery @ShawnKimble !

Issue #57, opened 2023-12-14
New PW firmware seems to break vitals - 23.44.0


dscoffe39:
I like to begin by saying thank you for everyone's time, energy, and efforts toward this project; its been phenomenal to learn and use!

At 11:38pm last night I noticed that the pypowerwall was having authentication issues, I had to use the forget password function to reset the gateway password. Once I restarted everything I noticed that pypowerwall was no longer getting vitals and that the firmware version seems to be new.

output:
```
Device Version:
 23.44.0 eb113390

Device Vitals:
 None
```

generated by:
```
import pypowerwall

password='xxx'
email='xxx'
host = "xxx"
timezone = "America/Los_Angeles" 

pw = pypowerwall.Powerwall(host,password,email,timezone)

print("Device Version:\n %s\n" % pw.version())
print("Device Vitals:\n %s\n" % pw.vitals(True))
```

jasonacox:
Thanks for opening this @ganeshtheju .  If you can, please try some of these:

* If you are able to log in to the Gateway portal, can you provide a screenshot? 
* Try getting raw vitals to see if this is a protobuf update: `pw.poll('/api/devices/vitals')`
* Try some of the other APIs to see if there are any other changes.

ganeshtheju:
![Screenshot from 2023-12-14 22-29-36](https://github.com/jasonacox/pypowerwall/assets/6383225/8475f363-7625-422c-b051-56f99df0ebcb)


raw vitals don't work either: (both from pypowerwall, or directly trying to hit the teg vitals endpoint [returns a 404])

`Device Version:
 23.44.0 eb113390

Device Vitals:
 None

pw.poll:
 None
`

As far as i can tell, the pw charge/capacity/temps and alerts arent working, but everything else (meters etc) are working.

ganeshtheju:
It appears that there is atleast one other pypowerwall user who had gotten this update and is no longer getting vitals:
https://teslamotorsclub.com/tmc/posts/7965048/

dscoffe39:
It looks like the teg webgui gets its reserve values from /api/operation; even logged in as an installer I can't find any endpoints with the pod values or the alerts.

jasonacox:
@dscoffe39 Thanks!  If the /api/device/vitals API is no longer available (404) in this firmware release, that would explain why the other functions (PW charge, etc) are also missing.  Are you still seeing the power flow animation when you log in to the Gateway portal or use the Powerwall Dashboard?

[...]

PR #63, opened 2024-01-03
Add Bearer Token Auth

jasonacox:
This release adds the ability to use a Bearer Token for auth for local Powerwall gateway API calls. This is selectable by defining `authmode='token'` in the initialization. The default mode uses the existing `AuthCookie` and `UserRecord` method.

```python
import pypowerwall

pw = pypowerwall.Powerwall(HOST, PASSWORD, EMAIL, TIMEZONE, authmode="token")
```

This could potentially help with Issue #57 (untested)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ and call function to poll data.  Here is an example:
 
  Classes
     Powerwall(host, password, email, timezone, pwcacheexpire, timeout, poolmaxsize, 
-              cloudmode, siteid, authpath)
+              cloudmode, siteid, authpath, authmode)
 
  Parameters
     host                      # Hostname or IP of the Tesla gateway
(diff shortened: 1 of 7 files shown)


mcbirse:
Hi Jason - is it okay if I push some unrelated minor tweaks for pypowerwall as part of this PR before it gets merge? Just a thought to reduce the number of updates that need to pushed to PyPI and Docker Hub.

Mostly very minor - just to neaten up the cloud mode setup output...

mcbirse:
Oh, I did also have a thought to make a change to the "scan" option however, so that it could accept an additional argument - that being able to pass the users current host IP address.

I was thinking offering a "scan" of the users network during the setup process might be good... but as that would be run using "docker exec" and inside the container, it does not detect the correct network. Could be resolved easily by passing the current IP address of the host from the setup script and base the network detection on that instead.

jasonacox:
> is it okay if I push some unrelated minor tweaks for pypowerwall as part of this PR 

Absolutely! Please do.

> change to the "scan" option ... offering a "scan" of the users network during the setup process 

That's brilliant.  Yes please. I never thought of that but you are right, it would be a nice addition.

mcbirse:
Hope that small addition is okay...

I was looking at the setup script changes for Powerwall-Dashboard, and we could now include something like this to run a network scan:

```bash
docker exec -it pypowerwall python -m pypowerwall scan -ip=$(ip route get 8.8.8.8 | awk '{ print $NF }')
```

jasonacox:
v0.7.4 Merged and uploaded:

* https://pypi.org/project/pypowerwall/0.7.4/
* jasonacox/pypowerwall:0.7.4t38

PR #67, opened 2024-01-21
v0.7.7 - Battery Data and Network Scanner

jasonacox:
## Proxy v0.7.7 t40 - Battery Vitals Update

* Proxy: Use /api/system_status battery blocks data to augment /pod and /freq macro data APIs. 

This update will help address some of the data lost with the removal of the `vitals` data starting with Firmware 23.44.0 as raised in https://github.com/jasonacox/Powerwall-Dashboard/discussions/402. 

## pyPowerwall v0.7.7 - Network Scanner Update

* Network Scanner: Improve network scan speed by scanning multiple hosts simultaneously. The number of hosts to scan simultaneously can be adjusted using the optional `-hosts=` argument (default = 30, maximum = 100), e.g. `python -m pypowerwall scan -hosts=50`

### DIFF
--- a/proxy/server.py
+++ b/proxy/server.py
@@ -265,7 +265,25 @@ class handler(BaseHTTPRequestHandler):
             # Frequency, Current, Voltage and Grid Status
             fcv = {}
             idx = 1
+            # Pull freq, current, voltage of each Powerwall via system_status
+            d = pw.system_status() or {}
+            if "battery_blocks" in d:
+                for block in d["battery_blocks"]:
+                    fcv["PW%d_name" % idx] = None # Placeholder for vitals
+                    fcv["PW%d_PINV_Fout" % idx] = get_value(block, "f_out")
+                    fcv["PW%d_PINV_VSplit1" % idx] = None # Placeholder for vitals
+                    fcv["PW%d_PINV_VSplit2" % idx] = None # Placeholder for vitals
+                    fcv["PW%d_PackagePartNumber" % idx] = get_value(block, "PackagePartNumber")
+                    fcv["PW%d_PackageSerialNumber" % idx] = get_value(block, "PackageSerialNumber")
+                    fcv["PW%d_p_out" % idx] = get_value(block, "p_out")
+                    fcv["PW%d_q_out" % idx] = get_value(block, "q_out")
+                    fcv["PW%d_v_out" % idx] = get_value(block, "v_out")
+                    fcv["PW%d_f_out" % idx] = get_value(block, "f_out")
+                    fcv["PW%d_i_out" % idx] = get_value(block, "i_out")
+                    idx = idx + 1
+            # Pull freq, current, voltage of each Powerwall via vitals if available
             vitals = pw.vitals() or {}
+            idx = 1
             for device in vitals:
                 d = vitals[device]
                 if  device.startswith('TEPINV'):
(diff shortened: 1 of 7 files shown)


mcbirse:
Before merging would you be interested in a change to the network scanning?

I just did a quick hack/test making it multi-threaded, and using no more than 10 threads at a time it scans my entire network in a faction of the time...

jasonacox:
Yes please @mcbirse !  Can you merge it into this branch / PR?

mcbirse:
Will do - it's a bit raw at the moment so I will cleanup/refine before committing.

I literally just spent the last 20mins seeing if it would work, so have not spent much time on the code/testing yet. And I may have a bug where it actually created 255 threads simultaneously. It sure got a result quickly! 🤣

mcbirse:
Hmm no it was 10 threads only. Wow - the improvement to scanning speed this makes is amazing!

I had noted your comment about the slow scanning, and had noticed this before myself as well with a thought to make it multi-threaded.

Just need to add some options, to e.g. control the max number of threads etc. Stay tuned.

mcbirse:
Added the changes for simultaneously scanning multiple hosts with the network scan using threads.

I was testing and trying to determine a sensible default max number of threads?

I've gone with 32 - it seems like a lot, but researching various network scan tools out there it actually it seems reasonable. I found others with about 30 as a default, another with 64 (Angry IP scanner)... and generally more than 10 is common.

On Windows the scan is slow, so a larger number of threads reduces the overall scan time. On Linux it will be lightning fast to scan your whole subnet (depending on how many actual live hosts you have on the network).

Please test - happy to adjust defaults. I have not tested impact/load on system resources due to the multiple threads.

mcbirse:
TODO: I will review the "Battery Vitals Update" changes next - sorry for hijacking this PR with my changes! 😄

mcbirse:
> I've gone with 32

I may change this to 30 and set a built in max of 100. Found several other network scan tools that seem to use these numbers... not sure why but I assume they know better than I do.

pypowerwall/scan.py
```diff
@@ -35,7 +48,50 @@ def getmyIP():
...
+        # Disable Terminal Color Formatting
+        bold = subbold = normal = dim = alert = alertdim = ""
+
+    host = dim + '\r      Host: ' + subbold + '%s ...' % addr + normal
+    print(host, end='')
+    a_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    a_socket.settimeout(timeout)
+    location = (str(addr), 443)
+    result_of_check = a_socket.connect_ex(location)
+    if result_of_check == 0:
```
jasonacox, line 65: I found a pre-existing bug! Your change is lighting fast, but I noticed that every once in a while it wouldn't find my Powerwall. I have a LOT of devices on my network including several with port 443 open and the number it found would vary. 

I looked it up and connect_ex() can return errno.EAGAIN which indicates that some blocking event occurred so instead of waiting, it just hands back errno.EAGAIN.  I add a print(result_of_check) and saw several EAGAIN errors. I made a test change to essential honor errno.EAGAIN by "trying again" after minor delay:

```python
    while True:
        result_of_check = a_socket.connect_ex(location)
        if not result_of_check == socket.errno.EAGAIN:
            break
        time.sleep(0.1)
    if result_of_check == 0:
```

With that change, it gives consistent results. I tested on MacOS and Linux.  I need to check WinOS.

mcbirse: Great find!

proxy/server.py
jasonacox, line 1: Both system_status() and vitals() payloads were stored in `d`, causing aggregate data to invalid in cloud mode.

mcbirse:
Testing latest changes to Battery Data / Vitals successfully in both Local and Cloud mode.

Checked output of both /freq and /pod and looks good.

I am seeing the below error in pypowerall logs however (even in non-debug mode) - when in local mode

```
01/21/2024 06:32:35 PM [proxy] [ERROR] Missing key in payload [nominal_full_pack_energy]
01/21/2024 06:32:35 PM [proxy] [ERROR] Missing key in payload [nominal_energy_remaining]
```

Due to aggregated data now includes below and passes a null value to the `get_value` function.

```json
  "nominal_full_pack_energy": null,
  "nominal_energy_remaining": null,
```

I wonder if we should change/suppress this? Check for null before calling the function?

mcbirse:
I think this is bug - I should not be getting null values for this in either mode. Not sure why yet - will check later tonight if time.

```
  "nominal_full_pack_energy": null,
  "nominal_energy_remaining": null,
```

jasonacox:
> 01/21/2024 06:32:35 PM [proxy] [ERROR] Missing key in payload [nominal_full_pack_energy]
> 01/21/2024 06:32:35 PM [proxy] [ERROR] Missing key in payload [nominal_energy_remaining]

Yes, that was the bug I caught where vitals() and system_status() were using the same variable. Try with this commit:
https://github.com/jasonacox/pypowerwall/pull/67/commits/2da87131eaffec6b089aea2b9f1d7a8301614a50

mcbirse:
I will re-test... I was sure I pulled that commit!! But very likely I have stuffed up with my test setup here - sorry!

mcbirse:
All good - my mistake, I had not pulled the latest changes! 😕 

```json
{
  "backup_reserve_percent": 30,
  "nominal_full_pack_energy": 13051,
  "nominal_energy_remaining": 7334.157894736843,
  "time_remaining_hours": 1.5568100604639636
}
```

mcbirse (approved):

jasonacox:
No worries! I'm glad you saw the same thing I did! If you have time, feel free to merge. I'll push to PyPI later today unless we find anything else.

jasonacox:
Pushed:
* https://pypi.org/project/pypowerwall/0.7.7/
* jasonacox/pypowerwall:0.7.7t40

Thanks @mcbirse !

PR #72, opened 2024-02-24
ignore sites that are not complete set up

dcgibbons:
While Tesla is trying to figure out how to properly set up my two-gateway and three-powerwall system, they created a blank Site on my account with no devices attach. In this scenario pypowerwall will crash trying to display information about sites during the site selection menu creation.

This patch just skips those if that fails. I suspect you might want another way of fixing it, but sharing this for discussion. It does work in the case I'm seeing right now (I have 2 sites at the moment, including this bogus one).

Photo of Tesla app for context attached.

IMG_6870

### DIFF
--- a/pypowerwall/cloud.py
+++ b/pypowerwall/cloud.py
@@ -906,8 +906,11 @@ class TeslaCloud:
             else:
                 sitelabel = " "
             siteids.append(s["energy_site_id"])
-            print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, s["site_name"], 
+            try:
+                print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, s["site_name"],
                     s["energy_site_id"], s["resource_type"]))
+            except KeyError:
+                pass
             idx += 1
         # Ask user to select a site
         while True:
--- a/pypowerwall/cloud.py
+++ b/pypowerwall/cloud.py
@@ -906,11 +906,12 @@ class TeslaCloud:
             else:
                 sitelabel = " "
             siteids.append(s["energy_site_id"])
-            try:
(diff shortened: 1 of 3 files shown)


pypowerwall/cloud.py
```diff
@@ -906,8 +906,11 @@ def setup(self, email=None):
             else:
                 sitelabel = " "
             siteids.append(s["energy_site_id"])
-            print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, s["site_name"], 
+            try:
+                print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, s["site_name"],
                     s["energy_site_id"], s["resource_type"]))
+            except KeyError:
+                pass
```
jasonacox, line 913: Thanks for this @dcgibbons !

Can you check to see if this would work instead of the try/except block?

```python
            if "site_name" in s and "resource_type" in s:
                print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, s["site_name"],
                    s["energy_site_id"], s["resource_type"]))
            else:
                print(" %s%d - %s (%s) - Type: %s" % (sitelabel, idx, "Unknown",
                    s["energy_site_id"], "Unknown"))
```

That way it prints a placeholder for this bad site location and also keeps a consistent index for the user to select (versus skipping numbers).

dcgibbons: Yup! That works and I updated this PR with that.

```
2 Sites Found (* = default)
 *1 - 200 Amp Panel (XXXXXX) - Type: battery
  2 - Unknown (YYYYYY) - Type: Unknown

  Select a site [1]: 1
```

jasonacox:
https://pypi.org/project/pypowerwall/0.7.8/

PR #75, opened 2024-03-17
Adding alternate alerts when vitals not present

DerickJohnson:
This will utilize the /api/solar_powerwall endpoint to populate the alerts array if devices (through vitals) are not present. Let me know if you want it simplified to not include the outer 'if'. I included it for clarity, to show the distinct paths, but it can also be simplified if needed (since it won't loop through devices that aren't present).

I also only used the fallback if alertsonly was true, since we don't have device specific information in that setup.

This should be automatically picked up by the alerts/pw endpoint since the format is the same.

I tested it locally using the /alerts endpoint, but might be good to validate that it still uses vitals if they are present (it should). I am on the later firmware where it's no longer present (23.44). 

Let me know if you'd like anything changed!

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,29 @@
 # RELEASE NOTES
 
+## v0.7.12 - Cachefile and Alerts
+
+* Added alerts from `/api/solar_powerwall` when vitals not present by @DerickJohnson in #75. The vitals API is not present in firmware versions > 23.44, this provides a workaround to get alerts.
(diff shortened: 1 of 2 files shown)


DerickJohnson:
Let me add some error handling actually, in case that endpoint goes

DerickJohnson:
Ok, ready to be reviewed!

jasonacox:
Looks good @DerickJohnson !  Thanks! 🙏 

I'll add release notes and start the merge.

PR #76, opened 2024-03-17
Use solar_powerwall for string data

jasonacox:
Adds logic to pull string data from `/api/solar_powerwall` API if vitals data is not available.

This applies to Firmware > 23.44 and owners of Powerwall+ systems (integrated inverters). The API seems to only return the string attached tot he first Powerwall+ system. This will be changed out if we discover a way to get all strings.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,11 +1,9 @@
 # RELEASE NOTES
 
-## v0.7.13 - Strings
+## v0.7.12 - Cachefile and Vitals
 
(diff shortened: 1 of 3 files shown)


niabassey:
It's fantastic to have per string inverter data back!  And finally a complete list of PVS_a0** and PVAC_a0** Alerts.

Testing it now, and seems to be working well.

jasonacox:
Thanks for the update @niabassey !

PR #78, opened 2024-03-26
v0.8.1 - Set battery reserve, operation mode

emptywee:
* Added `get_mode()` function.
* Added `set_battery_op_reserve()` function to set battery operation mode and/or reserve level. Likely won't work in the local mode.
* Added basic validation for main class `__init__()` parameters (a.k.a. user input).


pypowerwall/local/pypowerwall_local.py
```diff
@@ -207,6 +207,65 @@ def poll(self, api: str, force: bool = False,
...
+        except requests.exceptions.ConnectionError:
+            log.debug('ERROR Unable to connect to Powerwall at %s' % url)
+            return None
+        except Exception as exc:
+            log.debug('ERROR Unknown error connecting to Powerwall at %s: %s' % (url, exc))
+            return None
+        if r.status_code == 404:
+            log.debug('404 Powerwall API not found at %s' % url)
+            return None
+        if 400 <= r.status_code < 500:
```
emptywee, line 237: In local mode any attempt to send payload to `/api/operation` results in 403. Should we treat it here as Unauthorized instead of re-attempting to re-login?

pypowerwall/__init__.py
```diff
@@ -606,3 +681,52 @@ def get_time_remaining(self) -> Optional[float]:
...
+        # If local mode, check appropriate parameters, too
+        else:
+            # Ensure we can create a cachefile
+            dirname = os.path.dirname(self.cachefile)
+            if not dirname:
+                dirname = '.'
+            self._check_if_dir_is_writable(dirname)
+
+    @staticmethod
+    def _check_if_dir_is_writable(dirpath):
```
emptywee, line 719: This needs to be tested on Windows, if it causes issues we can fix it or remove the whole thing or the particular piece of it which causes issues.

pypowerwall/__init__.py
```diff
@@ -475,6 +511,45 @@ def get_reserve(self, scale=True) -> Optional[float]:
...
+    def get_mode(self, force=False) -> Optional[float]:
+        """
+        Get Battery Operation Mode
+        """
+        data = self.poll('/api/operation', force=force)
+        if data and data.get('real_mode'):
+            return data['real_mode']
+        return None
+
+    def set_battery_op_reserve(self, level: Optional[float] = None, mode: Optional[str] = None,
```
emptywee, line 523: Open to suggestions if the name of the function is too long/odd.

pypowerwall/local/pypowerwall_local.py
jasonacox: Yes! Good idea.

pypowerwall/__init__.py
jasonacox: It is fine, but we could abbreviate as `set_operation()` since that is the Tesla API and it describes what it does, but I would like to see two interfaces to this that align to their "get" equivalents:

```
set_reserve(level)
set_mode(mode)
```

pypowerwall/__init__.py
```diff
@@ -184,13 +197,36 @@ def poll(self, api='/api/site_info/site_name', jsonformat=False, raw=False, recu
...
+            din         = System DIN (ignored in Cloud mode)
+            raw         = If True, send raw data back (useful for binary responses, has no meaning in Cloud mode)
+            recursive   = If True, this is a recursive call and do not allow additional recursive calls
+        """
+        response = self.client.post(api, payload, din, recursive, raw)
+        if jsonformat:
+            try:
+                return json.dumps(response)
+            except JSONDecodeError:
+                log.error(f"Unable to dump response '{response}' as JSON. I know you asked for it, sorry.")
```
jasonacox, line 226: You're funny. 😂

jasonacox:
This is great @emptywee !!!!

> Added set_battery_op_reserve() function to set battery operation mode and/or reserve level. Likely won't work in the local mode.

This still allows us to run this as a standalone script (in cloud mode) to change these settings (e.g. cron jobs or embedded in other scrips). Technically (read: Jason hasn't tried) a user of the library can instantiate two instances, one for local, one for cloud and use that one for control commands.

Will a local mode user get enough of an error to indicate they need to swtich to cloud mode? We could add that as a log error.

emptywee:
> Will a local mode user get enough of an error to indicate they need to swtich to cloud mode? We could add that as a log error.

With some modifications to error handling, it can produce:
```
Testing set_battery_op_reserve()...
DEBUG:https://192.168.1.1:443 "GET /api/operation HTTP/1.1" 200 117
DEBUG:https://192.168.1.1:443 "POST /api/operation HTTP/1.1" 403 115
ERROR:403 Unauthorized by Powerwall API at https://192.168.1.1/api/operation - Endpoint disabled in this firmware or user lacks permission
Set battery operation mode to backup failed.
```

emptywee:
@jasonacox ok see https://github.com/jasonacox/pypowerwall/pull/78/commits/8d100faf4bfef47b386b954faf329c86ee61f8f1

emptywee:
* Added `set_mode()` function.
* Added `set_reserve()` function.
* Handle 401/403 errors from Powerwall separately in local mode.
* Handle 50x errors from Powerwall in local mode.

emptywee:
OK, now I am not sure, does PW respond with 403 in local mode when session has expired? I'd respond with 401, but who knows? Do you know for sure @jasonacox ? Because your simulator uses 403 to force the example script to "re-login". If PW uses 403 for real to indicate an expired session, I'll have to adjust error handling in this case.

jasonacox:
> OK, now I am not sure, does PW respond with 403 in local mode when session has expired? I'd respond with 401, but who knows? Do you know for sure @jasonacox ? Because your simulator uses 403 to force the example script to "re-login". If PW uses 403 for real to indicate an expired session, I'll have to adjust error handling in this case.

Ugh, you are right. That's been the problem. It bounced between 401 and 403 with previous Firmware versions which is why we eventually went with the range. I haven't check to see what the current version is doing, but I suggest we keep the range or figure out some other logic.

pypowerwall/local/pypowerwall_local.py
```diff
@@ -165,13 +165,13 @@ def poll(self, api: str, force: bool = False,
                 self.pwcachetime[api] = time.perf_counter() + 600
                 self.pwcache[api] = None
                 return None
-            if r.status_code == 429:
+            elif r.status_code == 429:
                 # Rate limited - Switch to cooldown mode for 5 minutes
                 self.pwcooldown = time.perf_counter() + 300
                 log.error('429 Rate limited by Powerwall API at %s - Activating 5 minute cooldown' % url)
-                # Serve up cached data if it exists
+                # Serve up cached data if it exists (@emptywee: this doesn't look like we serve up cached data here?)
```
jasonacox, line 172: This comment needs to be removed.  If we hit this, we are at the TTL for the cached objects which is what the user expects us to do. After that TTL we should send fresh data or return None indicating something bad happened or no connection.

emptywee:
Well, in case of token expiration, they respond with 401:
```
< HTTP/2 401 
< content-type: application/json
< set-cookie: AuthCookie=; Path=/; Max-Age=0
< set-cookie: UserRecord=; Path=/; Max-Age=0
< x-content-type-options: nosniff
< content-length: 38
< date: Wed, 27 Mar 2024 19:22:39 GMT
< 
* Connection #0 to host 192.168.102.64 left intact
{"code":401,"message":"Token Expired"}
```

Should we adjust the simulation server to mirror this behavior if we want to test how the library re-authenticates?

jasonacox:
> Should we adjust the simulation server to mirror this behavior if we want to test how the library re-authenticates?

Yes, I'm good with that.

emptywee:
> Yes, I'm good with that.

Would you be able to build a new docker image for the pwsim? It is used in PR checks as `image: jasonacox/pwsimulator` and I cannot update the image in Docker Hub. Unless we want to run the pwsim from the source code of the PR? For that we'll need to change the GitHub Actions workflow, I think.

emptywee:
Or maybe the workflow could build a new image and push it Docker Hub automatically each time or if pwsim code was changed, or something along these lines.

jasonacox:
> Or maybe the workflow could build a new image and push it Docker Hub automatically each time or if pwsim code was changed, or something along these lines.

I pushed the update for pwsimulator and it passes now. I've always wanted to investigate how to do the automated docker push, but haven't. I have a build script that does it with little effort and typically would only run this when we cut a release.

```bash
docker buildx build --no-cache --platform linux/amd64,linux/arm64,linux/arm/v7 --push -t jasonacox/pwsimulator:${VER} .
```

emptywee:
> > OK, now I am not sure, does PW respond with 403 in local mode when session has expired? I'd respond with 401, but who knows? Do you know for sure @jasonacox ? Because your simulator uses 403 to force the example script to "re-login". If PW uses 403 for real to indicate an expired session, I'll have to adjust error handling in this case.
> 
> Ugh, you are right. That's been the problem. It bounced between 401 and 403 with previous Firmware versions which is why we eventually went with the range. I haven't check to see what the current version is doing, but I suggest we keep the range or figure out some other logic.

I'll test with my pw today, see how it behaves. 

> > Or maybe the workflow could build a new image and push it Docker Hub automatically each time or if pwsim code was changed, or something along these lines.
> 
> I pushed the update for pwsimulator and it passes now. I've always wanted to investigate how to do the automated docker push, but haven't. I have a build script that does it with little effort and typically would only run this when we cut a release.
> 
> ```shell
> docker buildx build --no-cache --platform linux/amd64,linux/arm64,linux/arm/v7 --push -t jasonacox/pwsimulator:${VER} .
> ```

https://docs.github.com/en/actions/publishing-packages/publishing-docker-images this?

emptywee:
Nice. I can make some little change to pwsim to test the new gh workflow :)

emptywee:
@jasonacox hmm, do you need to enable the new workflow somewhere? I didn't see it kicked off on my push to pwsimulator... or it has to be in the main branch?

emptywee:
Oh, it ran in my forked repo =)

emptywee:
@jasonacox would you mind setting up a repository level variable IMAGE_REPOSITORY for your repo to `jasonacox`? It would be at https://github.com/jasonacox/pypowerwall/settings/variables/actions

jasonacox:
Done but not sure if I did it right... tests are still failing.

image

emptywee:
Weird, it didn't pick it up:
image

emptywee:
@jasonacox  Ok, it's working for us with `${{ github.repository_owner }}`. Let's leave it for now until we figure it out as a separate chore for the repo. What'd you say?

jasonacox:
Crazy.  Thanks. I suspect it has to do with not being on the main branch. Anyway, I'll start testing.

emptywee:
> Crazy. Thanks. I suspect it has to do with not being on the main branch. Anyway, I'll start testing.

I don't think so, it picked up the changes in the workflow, but it wouldn't resolve the reference to the variable somehow. It is not in my main branch either, but it works in my repository. No idea what's wrong with GH here.

jasonacox:
So far, so good. Some minor things...

During testing I noticed that we need to be on TeslaPy >= 24.0. Also, I noticed that calling get_reserve() it seems like force (bypass cache) is not being used:

```
>>> import pypowerwall
>>> pypowerwall.set_debug(True)
>>> pw = pypowerwall.Powerwall(host="",email=email)
>>> pw.get_reserve(force=True)
DEBUG: -- cloud: Request for /api/operation
DEBUG: -- cloud: Returning cached SITE_CONFIG data
21.0
```

I'll see if I can figure out why.

emptywee:
I can look into it too

jasonacox:
I added command line arguments for setting/getting mode and reserve. Super handy to test things:

```
$ python3 -m pypowerwall set -reserve 20
pyPowerwall [0.8.1] - Set Powerwall Mode and Power Levels

Setting Powerwall Reserve to 20
```

```
$ python3 -m pypowerwall get 
pyPowerwall [0.8.1] - Set Powerwall Mode and Power Levels

  Site           Tesla Energy Gateway
  Din            1232100-00-E--TG123456789ABC
  Mode           self_consumption
  Reserve        20.0
  Current        41.53570872880039
  Grid           0
  Home           730
  Battery        730
  Solar          0
```

jasonacox:
I created a test container of the Proxy: jasonacox/pypowerwall:0.8.1t51-beta

Problem is that the new validation code is keeping it from starting. I need to see why... the authpath should be set by the `PW_AUTH_PATH` variable.

```
03/31/2024 12:30:24 AM [proxy] [INFO] pyPowerwall [0.8.1] Proxy Server [t51] - HTTP Port 8675
03/31/2024 12:30:24 AM [proxy] [INFO] pyPowerwall Proxy Started
03/31/2024 12:30:24 AM [proxy] [ERROR] Directory '.' is not writable. Check permissions.
03/31/2024 12:30:24 AM [proxy] [ERROR] Fatal Error: Unable to connect. Please fix config and restart.
```

jasonacox:
Ok, I commented out the file system check and build a new beta container.  Now I get a 403 error crashing the proxy:

03/31/2024 12:43:08 AM [proxy] [INFO] pyPowerwall [0.8.1] Proxy Server [t51] - HTTP Port 8675
03/31/2024 12:43:08 AM [proxy] [INFO] pyPowerwall Proxy Started
03/31/2024 12:43:08 AM [proxy] [INFO] pyPowerwall Proxy Server - Local Mode
03/31/2024 12:43:08 AM [pypowerwall.local.pypowerwall_local] [ERROR] 403 Unauthorized by Powerwall API at https://10.0.1.28/api/site_info/site_name - Endpoint disabled in this firmware or user lacks permission
Traceback (most recent call last):
  File "/app/server.py", line 174, in <module>
    log.info("Connected to Energy Gateway %s (%s)" % (host, pw.site_name().strip()))
AttributeError: 'NoneType' object has no attribute 'strip'
```

I know we thought that we should only see 401 and not 403. But for some reason I'm getting 403's again.

**Comment by jasonacox** [comment]
Latest proxy to test: `jasonacox/pypowerwall:0.8.1t52-beta1`

@mcbirse you may want to try this one... our old friend "SystemConnectedToGrid" has been ressurected 😁 

image

**Comment by emptywee** [comment]
> I know we thought that we should only see 401 and not 403. But for some reason I'm getting 403's again.

Says a lot about tesla api devs hehe...

**Comment by emptywee** [comment]
> I commented out the file system check

Hmm, so the problem is only when running in a container? Might be something odd with the check when path is mounted from the host. I'll look into it, but it's not a blocker for now. I anticipated something like that to happen :-)

**Comment by jasonacox** [comment]
> > I commented out the file system check
> 
> Hmm, so the problem is only when running in a container? Might be something odd with the check when path is mounted from the host. I'll look into it, but it's not a blocker for now. I anticipated something like that to happen :-)

Actually, the issue was that the proxy was using the cachefile default (local directory .cachefile) which would not be writable. This has been an issue for a long time! It wouldn't surface because the service would just use the password to log in at restart. However, that creates more load on the Powerwall. I updated proxy.

@emptywee your init check code caught this!  Love that! ❤️ 

I added some additional details to the init check messages to help troubleshoot and identify issues. I kept hunting the authpath one but it was really the cachefile one.

**Comment by emptywee** [comment]
Ah, glad that you find it helpful and not irritating :) was a bit afraid of the latter!

**Comment by emptywee** [comment]
@jasonacox How's testing going? Ready to merge or not yet?

**Comment by jasonacox** [comment]
I haven't had time to troubleshoot why the `force` option is not working with the new functions. It may not be passing it along to the poll(). Do you have time to check?

```
>>> import pypowerwall
>>> pypowerwall.set_debug(True)
>>> pw = pypowerwall.Powerwall(host="",email=email)
>>> pw.get_reserve(force=True)
DEBUG: -- cloud: Request for /api/operation
DEBUG: -- cloud: Returning cached SITE_CONFIG data
21.0
```

**Comment by emptywee** [comment]
Oh, absolutely. Thought you figured it out already and I was sitting duck waiting for you to merge it :P

**Comment by jasonacox** [comment]
You're amazing!  Great find.  Yes, the main cloud use up to this point was "read only" to get the metric data which is delayed in the cloud anyway, so it hasn't surfaced as an issue. However, with your great work here, we now have "write" functions that could change state that would not show as changed until the cache TTL expires. That doesn't seem ideal. In fact, I wonder if `get_reserve()` and `get_mode()` should always be forced by default?

I am still good with merging this but will wait if you want to try to add the force feature first.

**Review by jasonacox — APPROVED** [review]

**Comment by emptywee** [comment]
Sure, adding the force flag should be trivial at this point. We can clear certain cache entries once "write" operation is complete.

**Comment by emptywee** [comment]
@jasonacox I was thinking something like this https://github.com/jasonacox/pypowerwall/pull/78/commits/c46b479b62fcdd3c419be8151d5b20c2cbcf5110

**Comment by jasonacox** [comment]
Nice @emptywee !  I'm testing...

**Comment by jasonacox** [comment]
Ok, this is odd.  When you set_reserve(), and then try get_reserve() it will return None from cache. Something is wrong with the cache logic. But, if you force=True, it does work.

```
>>> pw.set_reserve(20)
DEBUG: -- cloud: Request for /api/operation
DEBUG:Invoking 0 protected resource request hooks.
... truncated ...
>>> pw.get_reserve()
DEBUG: -- cloud: Request for /api/operation
DEBUG: -- cloud: Returning cached SITE_CONFIG data
>>>
>>> pw.get_reserve(force=True)
DEBUG: -- cloud: Request for /api/operation
DEBUG:Invoking 0 protected resource request hooks.
[... trim ...]
DEBUG: -- cloud: Retrieved SITE_CONFIG data
20.0
>>>
>>> pw.get_reserve()
DEBUG: -- cloud: Request for /api/operation
DEBUG: -- cloud: Returning cached SITE_CONFIG data
20.0
>>>
```

emptywee:
@jasonacox yeah it was a silly thing. Can you check now? 

Sorry it took that long, I was taking a shower :)

emptywee:
Calling it a night now, but it should work now as intended.

jasonacox:
It works!!  Thanks @emptywee !

I'm testing `jasonacox/pypowerwall:0.8.1t52-beta4` - should be ready for merge soon.

jasonacox:
✅ Tests look great. Thanks @emptywee !!!  Great enhancements.

emptywee:
Woot woot!

mcbirse:
Brilliant work guys! Looks like you've both got all the testing covered... I'll test the new container at some point.

Sorry I haven't had a chance to be involved and help with this - got a work project on that is currently sapping all my time and energy.

jasonacox:
No worries @mcbirse !  I understand that. :-) 

I'm going to cut a PR for Powerwall-Dashboard that includes our latest pypowerwall container. It adds a handful of new alerts and seems to be performing well. A few in the community have already started running the beta.

A surprise to me was how much GridServicesActive flapps. I don't know what that may indicate but find it interesting.

image

emptywee:
My stackstorm actions now got simplified :) woohoo!

mcbirse:
> A surprise to me was how much GridServicesActive flapps. I don't know what that may indicate but find it interesting.

I'm not 100% certain, however I was under the impression that GridServicesActive is supposed to show when your Powerwall is participating in a VPP event...

For it to be toggling so frequently however, could it be used to indicate something else as well, perhaps?

PR #107, opened 2024-07-21
Add computed voltage and current to aggregates from TEDAPI status data

jasonacox:
## v0.10.9- TEDAPI Voltage & Current

* Add computed voltage and current to `/api/meters/aggregates` using system data from TEDAPI status payload.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,8 +1,9 @@
 # RELEASE NOTES
 
-## v0.10.8 - TEDAPI Voltage & Current
+## v0.10.9 - TEDAPI Voltage & Current
 
-* Add computed voltage and current to `/api/meters/aggregates` from TEDAPI status data. 
+* Add computed voltage and current to `/api/meters/aggregates` from TEDAPI status data.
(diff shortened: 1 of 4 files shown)


jasonacox:
Proxy test: jasonacox/pypowerwall:0.10.9t63-beta

BuongiornoTexas:
@jasonacox - I think I've run into a wrinkle with this - at least assuming this is used for system vitals. I've only just upgraded to 4.4.6 (been busy) and while I can see the system vitals, the voltages being reported for solar, home and grid are around the 415V level, while the powerwall reports around 235V. I'm guessing the 415V is being calculated assuming 120V/phase? I'd expect all of the voltages to be around 230-240V locally. 

I did look around both here and in Powerwall-Dashboard, but didn't see any issues addressing this. If there  is a live discussion, could you point me at it. If not, would you prefer an issue raised here or under Powerwall-Dashboard?

jasonacox:
Yes, let's open an issue. Where do you see the bad voltages?  When you get a chance, send me the raw data you see for /api/meters/aggregates (e.g. http://localhost:8675/api/meters/aggregates)

PR #113, opened 2024-11-16
Remove Negative Solar Values [Option]

jasonacox:
# Proxy t65 - Negative Solar Value [Option]

* Add PW_NEG_SOLAR config option and logic to remove negative solar values for `/aggregates` and `/csv` APIs
* Update http://pypowerwall:8675/stats and http://pypowerwall:8675/help to show config data

Reference: 

https://github.com/jasonacox/Powerwall-Dashboard/issues/543#issuecomment-2480684273


jasonacox:
jasonacox/pypowerwall:0.11.1t65

PR #114, opened 2024-11-25
Add Powerwall Temps

jasonacox:
## v0.12.0 - Add Controller Data

* TEDAPI: Add `get_device_controller()` to get device data which includes Powerwall THC_AmbientTemp. 
* Updated `vitals()` to include Powerwall temperature data using new `get_device_controller()` function.
* Proxy Updated to t66 to include API response for /tedapi/controller.

Credit to @ygelfand (and [go-powerwall proxy](https://github.com/ygelfand/go-powerwall)) for discovery and reported in https://github.com/jasonacox/Powerwall-Dashboard/discussions/392#discussioncomment-11360474

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,11 @@
 # RELEASE NOTES
 
+## v0.11.2 - Add Controller Data
+
+* TEDAPI: Add `get_device_controller()` to get device data which includes Powerwall THC_AmbientTemp data. Credit to @ygelfand for discovery and reported in https://github.com/jasonacox/Powerwall-Dashboard/discussions/392#discussioncomment-11360474
+* Updated `vitals()` to include Powerwall temperature data. 
+* Proxy Updated to t66 to include API response for /tedapi/controller.
+
 ## v0.11.1 - PW3 and FleetAPI Bugfix
 
 * TEDAPI: Fix bug with activeAlerts logic causing errors on systems with multiple Powerwall 3's. Identified by @rmotapar in https://github.com/jasonacox/Powerwall-Dashboard/issues/387#issuecomment-2336431741 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,10 +1,12 @@
 # RELEASE NOTES
 
-## v0.11.2 - Add Controller Data
+## v0.12.0 - Add Controller Data
 
 * TEDAPI: Add `get_device_controller()` to get device data which includes Powerwall THC_AmbientTemp data. Credit to @ygelfand for discovery and reported in https://github.com/jasonacox/Powerwall-Dashboard/discussions/392#discussioncomment-11360474
 * Updated `vitals()` to include Powerwall temperature data. 
 * Proxy Updated to t66 to include API response for /tedapi/controller.
+* Remove Negative Solar Values [Option] by @jasonacox in https://github.com/jasonacox/pypowerwall/pull/113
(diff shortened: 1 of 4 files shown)


jasonacox:
Example:

image

Beta proxy container set up for testing: jasonacox/pypowerwall:0.11.2t66-beta2

http://pypowerwall:8675/temps

```json
{
"TETHC--2012170-25-E--TG1234567890AB": 16.60000000000001,
"TETHC--3012170-05-B--TG1234567890AC": 16.900000000000006
}
```

jasonacox:
image

jasonacox:
Updated to v0.12.0 to reflect to highlight the major change in pulling vitals data.

jasonacox:
Proxy jasonacox/pypowerwall:0.12.0t66

nhasan:
Installed latest version 4.6.0 but the temps graph says "no data."

jasonacox:
Hi @nhasan - It requires that you are using TEDAP in hybrid or full mode. You can check using `verify.sh` - see https://github.com/jasonacox/Powerwall-Dashboard/pull/549#issuecomment-2509144281.

maxEntropyProd:
Great to see the PW temps back!! Great job!! Just a couple of notes that are mostly due to my lameness, but it might help others.

1. When I first ran add_route.sh, I entered the gateway IP from my local network, 192.168.0.xxx; however, i needed to use the gateway default wifi network of 192.168.91.1 This seems obvious in hindsight, but it might help others if that was the default (or just say use 192.168.91.1?)
2. When I then ran setup.sh, i did not change the PW credentials, so it did not prompt me for the pw for the Gateway, which is on the sticker of the inverter for PW+ (i.e., not the tesla pw). Consequently, TEDAPI showed as available, but not connected under verify.sh. After further reading here, i reran setup.sh, then said to update the credentials, which then ask me for the Gateway pw.

Question, it seems that pypowerwall is only using 192.168.91.1 and nothing from the local lan connection on 192.168.0.xxx. I assume that's correct, yes?

jasonacox:
> Question, it seems that pypowerwall is only using 192.168.91.1 and nothing from the local lan connection on 192.168.0.xxx. I assume that's correct, yes?

That is correct. For the PW2, all the old APIs are available on the 192.168.91.1 endpoint, but the TEDAPI data is there as well.

A note on "hybrid" versus "full" mode, for anyone curious: If verify.sh or the logs (`docker logs pypowerwall`) show it is running in hybrid mode, it is using the old APIs and the new TEDAPI both, which currently has the benefit of pulling a few more extended metrics. If it shows "full" mode, it is just using the TEDPAI endpoint and performs calculations to derive some of the extended metrics (e.g. line-to-line voltages, PV dc power, nom energy to be charged) and some is missing (eg. PV state). In other words, there is very little benefit to running hybrid but I kept it for consistency. To help simplify setup, I may remove it in a future release. As a note, hybrid is not available for Powerwall 3.

Thanks for the great suggestions! I'll see what we can do to help update the docs/setup. TODO: docs

jasonacox:
> Is there any way to change dashboard to show powerwall temps in Fahrenheit?

Yes, you can edit the panel and update the queries for all the Powerwalls with something like this:

Select the pencil (edit) button to switch from visual editor mode to query edit mode:

image


```sql
SELECT
  mean("PW1_temp") * 9 / 5 + 32
FROM "pwtemps"."http"
WHERE $timeFilter
GROUP BY time($__interval)
fill(null)
```

Edit the units to switch to Fahrenheit

image

PR #115, opened 2024-11-26
Fix errors with site references

Nexarian:
New Tesla user here, I have only Tesla solar inverters, and I was unable to access them with pypowerwall unless I applied this fix. Pypowerwall would crash when trying to find the site_name. The broken line was `self.siteid = sites[self.siteindex].get('energy_site_id')` in `pypowerwall_fleetapi.py`

Otherwise, make the change_site functions consistent and less error-prone! The change site function, especially the site name reference, can break users that only have Solar-Only Tesla inverters. Make the two change_site functions consistent and fix other issues where site_name is not provided by the Tesla API.


jasonacox:
Thanks @Nexarian !  Good fix to expand use for Solar-Only owners. I've merged and tested.  I'll include this in the upcoming release (v0.11.2).

PR #118, opened 2024-12-02
Improve route documentation and add example

Nexarian:
When using the TEDAPI over a LAN from Linux, it's critical to add a network route. However, I wanted to do this programmatically from Python before calling pypowerwall to get data.

This adds what I've learned and two example functions showing how to do just that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,18 +74,33 @@ The unofficial Tesla Owners API allows FleetAPI access (option 2) without having
 
 With version v0.10.0+, pypowerwall can access the TEDAPI endpoint on the Gateway. This API offers up additional metrics related to string data, voltages and alerts. However, you will need the Gateway/WiFi Password (often found on the QR sticker on the Powerwall Gateway). Additionally, your computer will need network access to the Gateway IP (192.168.91.1). You can have your computer join the Gateway local WiFi or you can add a route:
 
+In the examples below, change **192.168.0.100** to the IP address of Powerwall Gateway (or Inverter) on your LAN. Also, the **onlink** parameter may be necessary for Linux.
+
+#### Linux Ubuntu and RPi
 ```bash
-# Example - Change 192.168.0.100 to the IP address of Powerwall Gateway on your LAN
+# Can add to /etc/rc.local for persistence
+sudo ip route add 192.168.91.1 via 192.168.0.100 [onlink]
+```
 
-# Linux Ubuntu and RPi - Can add to /etc/rc.local for persistence 
-sudo ip route add 192.168.91.1 via 192.168.0.100
+See `examples/network_route.py` for two different approaches to do this programmatically in Python.
 
-# MacOS 
+#### MacOS
+```
 sudo route add -host 192.168.91.1 192.168.0.100 # Temporary 
 networksetup -setadditionalroutes Wi-Fi 192.168.91.1 255.255.255.255 192.168.0.100 # Persistent
+```
 
-# Windows - Using persistence flag - Administrator Shell
+#### Windows - Using persistence flag - Administrator Shell
+```
 route -p add 192.168.91.1 mask 255.255.255.255 192.168.0.100
+```
+
+#### Windows Subsystem For Linux (Version 2 specific)
+Follow the instructions for Linux, but you must edit (From the host Windows OS) `%USERPROFILE%\.wslconfig` and add the following settings:
+```
+[wsl2]
+networkingMode=mirrored
+```
 
 # Test
 python3 -m pypowerwall tedapi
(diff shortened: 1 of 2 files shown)


jasonacox:
Thanks @Nexarian !

You have a good point about making it easier for people to add the route. A contributor on the [Powerwall Dashboard](https://github.com/jasonacox/Powerwall-Dashboard/) project provided this script that adds the route for multiple OS variants. It may make more sense to just add that here. 

Here is the script: https://github.com/jasonacox/Powerwall-Dashboard/blob/main/add_route.sh

Your example is great for Linux. A significant portion of the community is using MacOS and Windows.  I'm good with it as an example we can include and reference, but we should also mention add_route.sh option or add support for the other OS variants.

Nexarian:
1. I'm surprised your documentation did not include `onlink` -- That is required for my use. `Onlink` appears to be a networking version of `sudo` or `force`. The OS simply assumes the route exists without validating. I couldn't find easy equivalents on Windows or Mac OS, but I didn't search very hard.

2. I am currently using WSL2. The `add_route.sh` script says that doesn't work. It does, I've included instructions in the PR that show how. You have to have a bridged network.

3. Good point about people using it for Mac OS & Linux. I haven't tested for either of those. Right now, I'm still trying to get Powerwall Dashboard to work at all on my 2x standalone inverters with a goal of running it from my Raspberry PI. Because I also want string data, I have to find a way to add route 1, consume data, delete route 1, add route 2, consume data, delete route 2, repeat just for that portion of the data. I'm not yet sure how to do that. I'm also trying to figure out how to redirect the influxdb storage to an external platter drive, as those are better for persistent read/write storage than a PI SD card (I know that time-series databases burn up SD cards).

4. We should pick where these instructions live, and then put ALL the data about it there and simply link to it. If it's more useful for Powerwall Dashboard, great. I see no reason why we can't have examples in both Python and bash.

You may also find this reddit thread useful: https://www.reddit.com/r/TeslaSolar/comments/1d4vpkg/teg_network_access/

jasonacox:
> I'm surprised your documentation did not include onlink

I'm not familiar with onlink and couldn't find anything in a search. Do you have more info? In any case, we would want a solution that works for the main supported platforms at least for the Dashboard project (Linux, MacOS, Win, RPi).

>  am currently using WSL2. The add_route.sh script says that doesn't work. It does

I'm not a Windows guru, so I probably didn't fully test this but the community reported that route changes in WSL do not stick (post reboot) and to get a persistent route set on the machine, the recommendation was to use the windows cmd `route -p add ${TARGET_IP} mask ${NETMASK} ${PW_IP}`. If there is a a better way, it may be good to add that.

> We should pick where these instructions live, and then put ALL the data about it there and simply link to it. If it's more useful for Powerwall Dashboard, great. I see no reason why we can't have examples in both Python and bash.

I'm not sure I fully understand. I suspect you have a brilliant idea here. Technically this is the repo for the python pypowerwall library that builds the pip package. I think this is the best place to put the python examples for routing and link to the bash option on the Dashboard page. Is that the idea?

Nexarian:
As to WSL: I need to do more testing, I'll get back to you. Also my Raspberry Pi 3 is just too slow (not enough RAM) to make this happen, so I've ordered a 5 to see if I can get things working on that.

Nexarian:
Ok I found a way to make both work simultaneously, at least on WSL. Again, ChatGPT helped :)

In my case inverter 1 is on 192.168.1.67 and inverter 2 is on 192.168.1.250 -- This configuration changes it so that when I curl their LAN addresses directly it just aliases to the internal one.

This avoids the awkward multiplexed switching that I had been doing.

```
echo "200 inverter1" | sudo tee -a /etc/iproute2/rt_tables
echo "201 inverter2" | sudo tee -a /etc/iproute2/rt_tables
```

Add routes to each inverter in their respective tables:

```
sudo ip route add 192.168.91.1 via 192.168.1.67 dev eth1 table inverter1
sudo ip route add 192.168.91.1 via 192.168.1.250 dev eth1 table inverter2
```

Use ip rule to direct traffic to the correct routing table:

```
sudo ip rule add from 192.168.1.67 lookup inverter1
sudo ip rule add from 192.168.1.250 lookup inverter2
```

This seemed to persist in spite of a reboot, likely because it creates a persistent route table!

Nexarian:
Now that I have the variation that does multiple TEGs, I'm looking at ways to implement it on Mac OS and Windows.

Nexarian:
@jasonacox A couple more things:
- I have improved versions of this now.
- I have done some investigating as to MacOS and Windows support... It's not always trivial to do that. There's a reason why most routers run Linux, its networking stack is simply superior. That being said, we can flag which things are Linux-only and ask for Windows or MacOS help for a port. pyroute2 for instance, only supports Linux.
- We still need to decide if the documentation for this problem will live in pypowerwall, Powerwall-Dashboard, or both.

For now; however, this is an improvement!

PR #129, opened 2025-01-04
Change GW_IP to self.gw_ip

Nexarian:
The hard-coded 192.168.91.1 for the Tesla Inverters internal endpoint doesn't always work if you're using NAT. Change this to be compatible with the rest of the file and to enable this use-case.

Tested with a custom NAT configuration to verify that the firmware loads properly with a non-192.168.91.1 host.

### DIFF
--- a/pypowerwall/tedapi/__init__.py
+++ b/pypowerwall/tedapi/__init__.py
@@ -493,7 +493,7 @@ class TEDAPI:
         pb.message.recipient.din = self.din  # DIN of Powerwall
         pb.message.firmware.request = ""
         pb.tail.value = 1
-        url = f'https://{GW_IP}/tedapi/v1'
+        url = f'https://{self.gw_ip}/tedapi/v1'
         try:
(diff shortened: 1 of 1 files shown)


pypowerwall/tedapi/__init__.py
```diff
@@ -493,7 +493,7 @@ def get_firmware_version(self, force=False, details=False):
         pb.message.recipient.din = self.din  # DIN of Powerwall
         pb.message.firmware.request = ""
         pb.tail.value = 1
-        url = f'https://{GW_IP}/tedapi/v1'
+        url = f'https://{self.gw_ip}/tedapi/v1'
```
jasonacox, line 496: Wow! Great fix. This should have been self.gw_ip.

jasonacox:
Thank you @Nexarian !  This looks good.

jasonacox:
https://pypi.org/project/pypowerwall/0.12.3/

PR #134, opened 2025-01-21
Update proxy for v0.12.3 and add /csv/v2 API support

jasonacox:
Minor update to bump library version and provide expanded data for simple clients via `/csv/v2` API endpoint. 

* pyPowerwall v0.12.3 - Adds Custom GW IP for TEDAPI. 
* Add new API /csv/v2 which extends /csv by adding grids status (1/0) and battery reserve (%)setting:

```python
# Grid,Home,Solar,Battery,Battery_Level,Grid_Status,Reserve
-76.00,1114.25,360.00,780.00,96.51,1,20
```

### DIFF
--- a/proxy/RELEASE.md
+++ b/proxy/RELEASE.md
@@ -1,5 +1,14 @@
 ## pyPowerwall Proxy Release Notes
 
+### Proxy t68 (20 Jan 2025)
+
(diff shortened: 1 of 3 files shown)


jasonacox:
Container for testing: jasonacox/pypowerwall:0.12.3t68-beta2

PR #139, opened 2025-02-04
Improvements to grid status connected

Nexarian:
In certain TEDAPI edge cases the grid status was not correct. This fixes it so that "SystemConnectedToGrid" is the only alert that appears if the Tesla alerts API contains such a thing. It also removes risk of duplicates and normalizes this particular alert.

### DIFF
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -88,7 +88,7 @@ import time
 from json import JSONDecodeError
 from typing import Optional, Union
 
-version_tuple = (0, 12, 3)
+version_tuple = (0, 12, 4)
 version = __version__ = '%d.%d.%d' % version_tuple
 __author__ = 'jasonacox'
 
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -614,10 +614,8 @@ class Powerwall(object):
             elif alert:
                 alerts.add(alert)
 
-        def normalize_alerts(s: str) -> str:
-            return s.replace("SystemGridConnected", "SystemConnectedToGrid")
-
-        normalized_alerts = {normalize_alerts(s) for s in alerts}
+        # In the future, if more replacements are needed, create a utility function here.
+        normalized_alerts = list({s.replace("SystemGridConnected", "SystemConnectedToGrid") for s in alerts})
 
         if jsonformat:
             return json.dumps(normalized_alerts, indent=4, sort_keys=True)
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -84,9 +84,9 @@ import json
 import logging
 import os.path
 import sys
-from json import JSONDecodeError
-from typing import Union, Optional
 import time
+from json import JSONDecodeError
+from typing import Optional, Union
 
 version_tuple = (0, 12, 3)
 version = __version__ = '%d.%d.%d' % version_tuple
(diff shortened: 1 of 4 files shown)


pypowerwall/__init__.py
```diff
@@ -581,45 +581,47 @@ def alerts(self, jsonformat=False, alertsonly=True) -> Union[list, str]:
...
+            return s.replace("SystemGridConnected", "SystemConnectedToGrid")
+
+        normalized_alerts = {normalize_alerts(s) for s in alerts}
 
         if jsonformat:
-            return json.dumps(alerts, indent=4, sort_keys=True)
-        else:
-            return alerts
+            return json.dumps(normalized_alerts, indent=4, sort_keys=True)
+        return normalized_alerts
```
jasonacox, line 624: I like the move to set() to remove duplicates.  However, this would be a breaking change since the current API returns a list.

* Before

['PVS_a018_MciStringB', 'PVS_a059_MciOpen', 'SystemConnectedToGrid', 'FWUpdateSucceeded', 'GridCodesWrite', 'PodCommissionTime', 'PINV_a067_overvoltageNeutralChassis', 'PINV_a067_overvoltageNeutralChassis', 'SystemGridConnected']

* After

{'GridCodesWrite', 'PodCommissionTime', 'FWUpdateSucceeded', 'PVS_a018_MciStringB', 'SystemConnectedToGrid', 'PINV_a067_overvoltageNeutralChassis', 'PVS_a059_MciOpen'}

* Fix?

```python
        if jsonformat:
            return json.dumps(list(normalized_alerts), indent=4, sort_keys=True)
        return list(normalized_alerts)
```

['PINV_a067_overvoltageNeutralChassis', 'PodCommissionTime', 'SystemConnectedToGrid', 'FWUpdateSucceeded', 'GridCodesWrite', 'PVS_a059_MciOpen', 'PVS_a018_MciStringB']

pypowerwall/__init__.py
```diff
@@ -581,45 +581,47 @@ def alerts(self, jsonformat=False, alertsonly=True) -> Union[list, str]:
...
             if grid_status.get('grid_services_active'):
-                alerts.append('GridServicesActive')
+                alerts.add('GridServicesActive')
+            elif alert:
+                alerts.add(alert)
+
+        def normalize_alerts(s: str) -> str:
+            return s.replace("SystemGridConnected", "SystemConnectedToGrid")
+
+        normalized_alerts = {normalize_alerts(s) for s in alerts}
```
jasonacox, line 620: Instead of creating a function, why not do this?

```python
        normalized_alerts = {s.replace("SystemGridConnected", "SystemConnectedToGrid") for s in alerts}
```

jasonacox:
Thanks for this @Nexarian ! Looks good. I suggested a few changes.

pypowerwall/__init__.py
Nexarian: Because I wanted to leave room for more substitutions in the future. However, it is premature optimization and does increase the complexity of the function needlessly, so I'll keep it simple.

pypowerwall/__init__.py
Nexarian: Just transform normalized alerts to list when creating it:

`normalized_alerts = list({s.replace("SystemGridConnected", "SystemConnectedToGrid") for s in alerts})`

jasonacox:
Testing: jasonacox/pypowerwall:0.12.4t67-beta

jasonacox:
image

Showing alerts persisting after update and removal of redundant "SystemGridConnected" alert.

jasonacox (approved):
Looks good. Tested - no issues seen.

PR #143, opened 2025-02-10
Refactoring the proxy.

Nexarian:
- Enable ability to handle multiple gateways.

Fixing aggregates and problems with grid status

Improvements in accuracy and functionality

Changing alerts to be consistent with the -3 suffix


proxy/web/example.html
```diff
@@ -16,7 +16,7 @@ <h2 style="color:gray;font-family:verdana;">iFrame Example</h2>
     <script>
         // Get Firmware Version
         function showversion() {
-            var pwver = window.location.protocol + "//" + window.location.hostname + ":8675/version";
+            var pwver = window.location.protocol + "//" + window.location.hostname + ":" + window.location.port + "/version";
```
jasonacox, line 19: Yes! This needed to be some time ago. Thank you!

proxy/Dockerfile
```diff
@@ -1,7 +1,12 @@
-FROM python:3.10-alpine
+FROM python:3.13-alpine
 WORKDIR /app
-COPY requirements.txt /app/requirements.txt
+COPY ./proxy/requirements.txt /app/requirements.txt
```
jasonacox, line 3: I assume most would build the container from the directory that contains the Dockerfile (proxy), which is where my automation goes to build and push the proxy container. This would break that. Did you have another idea?

Nexarian: I agree we should leave this alone. I did this because I needed to test changes to the proxy AND pypowerwall itself and I didn't know of another good way to build them together, did I miss the memo on how to do that?

Anyway, I think what I should do is move this configuration to a different Dockerfile and leave this one alone.

proxy/web/example.html
Nexarian: I still occasionally have issues with it loading and/or reporting correct data, but I feel like that can be solved incrementally and later.

proxy/Dockerfile
jasonacox: Fair point. My dev and beta build script copies the library before running the build (`cp -r ../pypowerwall .`)

proxy/web/example.html
jasonacox: The change causes the issue or is this something else? Any odd errors in the javascript console?

Nexarian: - Sometimes [this code path](https://github.com/jasonacox/pypowerwall/blob/16b1a4e5be326439614c586fd18cd9dcc5b0558e/proxy/server.py#L731) gets executed when `client.session` (Or really, the `session` variable) is `None`. I blocked that with a simple if check, but I'm not sure what `session` is supposed to be in the context of TEDAPI or what this code path is for. Maybe blocking it is indeed the best path.

- Also I've noticed that the first time after server startup, the widget sometimes doesn't load. I have to refresh a few times, I haven't solved that either.

- Finally, and I think this is more a Tesla bug we have to work around: In the case where my panels generate significantly more power than the house is using (and thus sending a large amount back to the grid) the house shows as a negative load with a warning icon next to it. This is wrong, the house is still using power, my inverters setup is bugged. My installer (and therefore by proxy, Tesla) say that it's a bug with Tesla's monitoring, but they haven't provided any information on when a potential fix may be coming.

I don't believe any are directly related to my refactor; however. Likely just exposed through testing.

Nexarian: image

But my house is actually using 1.7 kWh!

mccahan: > Also I've noticed that the first time after server startup, the widget sometimes doesn't load.

The related issue I had there was that the step to [connect to the Powerwall](https://github.com/jasonacox/pypowerwall/blob/main/proxy/server.py#L198-L203) is synchronous, and has to be completed before the web server starts serving. If that runs slow, sometimes you try to load the widget before the server exists, so you get nothin' and have to refresh.

I had a branch where I had added an `autoconnect` parameter in there, shifted the connect process to a subthread, and had the [web request handler](https://github.com/jasonacox/pypowerwall/blob/main/proxy/server.py#L266) just throw out a simple HTML page with a meta refresh in it so it would reload itself for when the connection was done, but I ditched it.

PR #145, opened 2025-02-16
Alerts in extract_grid_status can be None. Block this edge case.

Nexarian:
While doing some testing I saw this error:

```
    if "SystemConnectedToGrid" in alerts:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
```

I'm not sure what triggered it, but it's an edge case that needs to be fixed.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,9 @@
 # RELEASE NOTES
 
+## v0.12.7 - SystemConnectedToGrid Fix
+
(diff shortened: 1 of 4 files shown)


jasonacox:
Good catch.

jasonacox:
Proxy jasonacox/pypowerwall:0.12.7t68-beta

PR #146, opened 2025-02-17
Add simulator test commands

mccahan:
Working on the power flow animation, I wanted an easy way to adjust the simulator parameters without having to stop/restart it. I added in some test commands, like `curl -k https://localhost/test/battery-percentage/100.0`(set the battery to 100%), and some sample scenarios like `curl -k http://localhost/test/scenario/nighttime-outage` (disconnected from grid, powering from the battery). There's also a basic UI for sending those commands at https://localhost/test/

For funsies I made an external UI for this (https://github.com/mccahan/pypowerwall-simulator-control), but not including that code here.

### DIFF
--- a/pwsimulator/README.md
+++ b/pwsimulator/README.md
@@ -78,3 +78,36 @@ docker stop pypowerwall
 # Start the server
 docker start pypowerwall
 ```
+
+## Test Commands
+
+### Battery
+Full: `curl -k https://localhost/test/battery-percentage/100.0`
+Empty: `curl -k https://localhost/test/battery-percentage/0.0`
+
+### Grid
+Toggle Grid Connection: `curl -k https://localhost/test/toggle-grid`
+
+### Solar
+Zero solar: `curl -k https://localhost/test/solar-power/0`
+Some solar: `curl -k https://localhost/test/solar-power/1450`
+
+### Scenarios
+This script includes some sample scenarios to cover common use cases
+
+```sh
+# Flow Scenarios
+curl -k http://localhost/test/scenario/battery-exporting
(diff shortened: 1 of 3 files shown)


jasonacox:
I love this!!!  Nice @mccahan !  

My test - and I think we should update the readme:

## Run Simulator and pyPowerwall

1. Run pwsimulator
```bash
cd pwsimulator; python stub.py
```
2. Run pypowerwall 
```bash
PW_HOST=localhost \
PW_PASSWORD=password \
PW_EMAIL=me@example.com \
PW_DEBUG=yes python3 proxy/server.py
```

## Simulator Control Panel

https://localhost/test/

image

https://localhost:8675/example.html

image

> For funsies I made an external UI for this (https://github.com/mccahan/pypowerwall-simulator-control), but not including that code here.

Excellent work on that! It would be nice if that was in the built-in UI (/test/) but I see it is in TS and would be a bit of work to port. At a minimum, we should add the instructions to run it in the README:

## Powerwall Scenario Simulator
```bash
docker run --rm -p 3000:3000 mccahan/pypowerwall-simulator-control:latest
```
image

jasonacox:
Future idea:

Ideally on the built-in control panel, it would be nice if it presented the internal state it was simulating. I think we would just need to expose:

```python
agg_solar = 6500
agg_home = 900
agg_grid = -2100
agg_powerwall = -3500
percentage = 23.975388097174584
active_scenario = ''
```

It would be nice to manipulate those as well, but we would have to add logic to ensure a zero balance and limits.

jasonacox:
Great work on this @mccahan !  Thanks for the addition.

Uploaded latest: jasonacox/pwsimulator:0.1.0

PR #149, opened 2025-03-03
Add option to get CSV headers

mccahan:
Playing around with some widget stuff, the CSV endpoints are very convenient but it's helpful to be able to get them with headers. I put it in as a query string and made sure it won't affect any existing URL calls. I switched the order around starting on line ~386 because otherwise `self.path.startswith('/csv')` would match `/csv/v2` as well

http://localhost:8675/csv/v2?headers
http://localhost:8675/csv?headers


jasonacox:
I like it!  Thanks @mccahan !

Running tests...

jasonacox:
Proxy: jasonacox/pypowerwall:0.12.7t69-beta

PR #150, opened 2025-03-13
Avoid divide by zero when nominalFullPackEnergyWh is zero

rlpm:
e.g. due to AC being disconnected

My PW3 was recently installed, but has not yet been granted PTO. So I have to keep it disconnected from AC.
In this state, nominalFullPackEnergyWh is zero, so `pypowerwall tedapi` fails with a divide by zero error.

### DIFF
--- a/pypowerwall/tedapi/__main__.py
+++ b/pypowerwall/tedapi/__main__.py
@@ -115,8 +115,11 @@ def run_tedapi_test(auto=False, debug=False):
     print(" - Power Data:")
     nominalEnergyRemainingWh = status.get('control', {}).get('systemStatus', {}).get('nominalEnergyRemainingWh', 0)
     nominalFullPackEnergyWh = status.get('control', {}).get('systemStatus', {}).get('nominalFullPackEnergyWh', 0)
-    soe = round(nominalEnergyRemainingWh / nominalFullPackEnergyWh * 100, 2)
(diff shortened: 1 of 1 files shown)


jasonacox:
Good find!

pypowerwall/tedapi/__main__.py
```diff
@@ -115,8 +115,11 @@ def set_debug(toggle=True, color=True):
     print(" - Power Data:")
     nominalEnergyRemainingWh = status.get('control', {}).get('systemStatus', {}).get('nominalEnergyRemainingWh', 0)
     nominalFullPackEnergyWh = status.get('control', {}).get('systemStatus', {}).get('nominalFullPackEnergyWh', 0)
-    soe = round(nominalEnergyRemainingWh / nominalFullPackEnergyWh * 100, 2)
-    print(f"   - Battery Charge: {soe}% ({nominalEnergyRemainingWh}Wh of {nominalFullPackEnergyWh}Wh)")
+    if nominalFullPackEnergyWh is 0:
```
jasonacox, line 118: We will want to change this to: `if nominalFullPackEnergyWh == 0:` 

This is because it checks for equality, which is the intended behavior. The proposed `if nominalFullPackEnergyWh is 0:`, checks for identity, i.e., whether the two variables are the same object in memory. This is not what we want to check in this case.

rlpm: `is` worked against my disconnected-from-AC PW3. But TIL it's because python caches small integer values.
Thanks for the note.

PR #152, opened 2025-03-16
Add fan speeds to TEDAPI monitoring

Nexarian:
# Summary
- Misc code cleanup in adjacent areas.
- Of note: No need to change the "code" parameter. The value field is apparently not validated against it.

# Testing
Works on my dual-inverter setup with 2x 7.6 kW Tesla inverters.
image

### DIFF
--- a/pypowerwall/tedapi/__init__.py
+++ b/pypowerwall/tedapi/__init__.py
@@ -51,7 +51,7 @@ import threading
 import time
 from functools import wraps
 from http import HTTPStatus
-from typing import List
+from typing import Dict, List
 
 import requests
 import urllib3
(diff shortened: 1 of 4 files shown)


jasonacox:
Thanks @Nexarian  - Great discovery and addition.

pypowerwall/tedapi/__init__.py
```diff
@@ -932,6 +932,28 @@ def battery_level(self, force=False):
         return battery_level
 
 
+    def extract_fan_speeds(self, data) -> Dict[str, Dict[str, str]]:
```
jasonacox, line 935: Shouldn't this helper function just go in inside `def vitalsI()` or prefix with `_` to identify it as an internal function?  Or better yet, the more I think about it, the more I like making it a SDK function call of TEDAPI, perhaps?

```python
def fan_speeds(self, force=False, data=None) -> Dict[str, Dict[str, str]]:
        fan_speed_signal_names = {"PVAC_Fan_Speed_Actual_RPM", "PVAC_Fan_Speed_Target_RPM"}

        # List to store the valid fan speed values
        result = {}

        # Get the device controller data if not provided
        if not data:
            data = self.get_device_controller(force=force)
...
```

Nexarian: I'm not sure I understand -- It's already inside the TEDAPI class?

Nexarian: @jasonacox Ping on my question here. I'd like to get this merged.

jasonacox: It's a minor thing, for anyone wanting to work directly with the tedapi class:

```python
import pypowerwall
pw = pypowerwall.Powerwall(PW_HOST, gw_pwd=PW_GW_PWD)

pw.tedapi.get_din()
pw.tedapi.get_firmware_version()

pw.tedapi.extract_fan_speeds(pw.tedapi.get_device_controller()) # using helper func

```

Hum.. I have an idea...

Nexarian: I like your idea :)

jasonacox:
https://pypi.org/project/pypowerwall/0.12.9/

asonacox/pypowerwall:0.12.9t69

PR #153, opened 2025-03-22
Refactor TEDAPI __init__ for readability/code reuse

mccahan:
This might be a bit much to bite off, so my feelings aren't hurt if we just skip it. I started just splitting the protobuf messages out to make the file more readable but then went on a yak-shaving expedition for a few things

Primary goal here is that the tedapi __init__ file is massive and has a lot of copy/paste repeated code in it. To that end:
- Let the `@uses_api_lock` decorator handle acquiring the locks
- Add `@uses_cache(key)` and `@uses_connection_required` decorators so each function call doesn't have to repeat that code
- Add a unified `__run_request` function to handle throttling, exceptions, etc
- Split the protobuf messages into their own classes with encode/decode functions, because they're massive and made the __init__ file so much more difficult to read
- Move the big chunks of the vitals dictionary to `vitals_dictionary.py` for readability

From my testing everything seems to be working, but I only have a single PW3 so might be missing hardware to thoroughly test some pieces.

@Nexarian this is probably pertinent to your interests


jasonacox:
Thanks @mccahan ! 

Yes, I agree, it is a lot but I don't hate it. 😄 I pulled it down and the basic functions work, but pulling vitals throws errors:

  File "/pypowerwall/tedapi/vitals_dictionary.py", line 221, in __get_pvs_and_pvac_strings
    pvac_strings[pvac_name]["PVAC_PvState_A"] = "PV_Active" if string_a else "PV_Disabled"
    ~~~~~~~~~~~~^^^^^^^^^^^
KeyError: 'PVAC--1538100-00-F--CN************J'

I didn't unwind the new vitals code so could be something simple.

Some things I consider for refactors like this:
* Is it as simple as we can make it for maintainability?
* Is it performant for tiny systems (e.g. RPi or NAS installs)?
* Are there edge cases (Powerwall systems vary a lot) that would break?

mccahan:
Theoretically, I may have fixed that issue - apparently I don't have whatever kind of hardware that is? I tested this by opening `/vitals` on my production instance side-by-side with this branch and it matched, so I'm assuming I don't have that.

- Personally I think it's easier to maintain when you've got things a little more modularized - finding things in that big file was a little dizzying, and having cache/locks as decorators feels like it simplifies
- Should be zero impact on tiny systems - there's really no new logic happening that wasn't happening already, and compared to the latency of waiting for responses from the PW itself anything we do here is just rearranging deck chairs on the titanic

pypowerwall/tedapi/__init__.py
```diff
@@ -1,7 +1,7 @@
 # pyPowerWall - Tesla TEDAPI Class
 # -*- coding: utf-8 -*-
 """
- Tesla TEADAPI Class
+ Tesla TEDAPI Class
```
Nexarian, line 4: Be still my heart. Fixing spelling errors!

pypowerwall/tedapi/__init__.py
```diff
@@ -95,9 +98,11 @@ def uses_api_lock(func):
         func.api_lock = threading.Lock()
     @wraps(func)
     def wrapper(*args, **kwargs):
-        # Inject the function object itself into kwargs.
-        kwargs['self_function'] = func
-        return func(*args, **kwargs)
+        # Wrap the function with the lock
+        self = args[0]
+        with acquire_lock_with_backoff(func, self.timeout):
```
Nexarian, line 103: If I understand correctly, this is a subtle change that wraps the ENTIRE function in the api lock, versus the specific parts of the function that are intended to be in a critical section. This reduces flexibility as there may be parts of each function that do not need to be locked.

pypowerwall/tedapi/decorators.py
```diff
@@ -1,5 +1,7 @@
 import functools
```
Nexarian, line 1: I have no objection to moving the api lock decorator here.

pypowerwall/tedapi/decorators.py
```diff
@@ -17,3 +19,57 @@ def wrapper(*args, **kwargs):
...
+# Cache Decorator
+# Checks to see whether a cached result exists for the function, and if so, returns it. If not, it calls the function and caches the result.
+def uses_cache(cache_key):
+    def decorator(func):
+        def wrapper(self, *args, **kwargs):
+            # We probably can't guarantee that "force" will be the first argument for the function, so let's find it
+            sig = inspect.signature(func)
+            namedArgs = sig.bind(self, *args, **kwargs).arguments
+            force = False
+            if 'force' in namedArgs:
```
Nexarian, line 46: Are we concerned with case sensitivity here?

pypowerwall/tedapi/decorators.py
```diff
@@ -17,3 +19,57 @@ def wrapper(*args, **kwargs):
...
+                force = namedArgs['force']
+
+            # Allow for dynamic cache keys using an argument to the function
+            # e.g. @uses_cache('battery-[args-din]') would cache the result of the function with the key 'battery-<din>'
+            use_cache_key = cache_key
+            if '[args-' in cache_key:
+                namedArg = cache_key.split('args-')[1].split(']')[0]
```
Nexarian, line 53: Would a regex be better to extract the values we want here?

pypowerwall/tedapi/exceptions.py
```diff
@@ -12,3 +12,11 @@ class PyPowerwallTEDAPINotImplemented(Exception):
 
 class PyPowerwallTEDAPIInvalidPayload(Exception):
     pass
+
+
+class PyPowerwallTEDAPIThrottleException(Exception):
```
Nexarian, line 17: Love these!

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
```diff
@@ -0,0 +1,41 @@
...
... signals(names: $pwsSignalNames) {\n        name\n        value\n        textValue\n        boolValue\n        timestamp\n      }\n      activeAlerts {\n        name\n      }\n    }\n    pch: components(filter: $pchComponentsFilter) {\n      signals(names: $pchSignalNames) {\n        name\n        value\n        textValue\n        boolValue\n        timestamp\n      }\n      activeAlerts {\n        name\n      }\n    }\n    bms: components(filter: $bmsComponentsFilter) {\n      signals(names: $bmsSignalNames) {\n        name\n        value\n        textValue\n        boolValue\n        timestamp\n      }\n      activeAlerts {\n        name\n      }\n    }\n    hvp: components(filter: $hvpComponentsFilter) {\n      partNumber\n      serialNumber\n      signals(names: $hvpSignalNames) {\n        name\n        value\n        textValue\n        boolValue\n        timestamp\n      }\n      activeAlerts {\n        name\n      }\n    }\n    baggr: components(filter: $baggrComponentsFilter) {\n      signals(names: $baggrSignalNames) {\n        name\n        value\n        textValue\n        boolValue\n        timestamp\n      }\n      activeAlerts {\n        name\n      }\n    }\n  }\n}\n"
```
Nexarian, line 25: General comment: Is it possible to break these out into more readable JSON so we can intelligently view what's truly in these queries?

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
+import json
```
Nexarian, line 1: "vitals_dictionary" is too verbose, you could just call it "vitals.py" -- It's already under TEDAPI, that's self-documenting enough.

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
...
+            "gateway": self.gw_ip,
+            "pyPowerwall": __version__,
+        }
+        return header
+
+    def get_meters(self):
+        meter_config = {}
+        if "meters" in self.config:
+            # Loop through each meter and use device_serial as the key
+            for meter in self.config['meters']:
```
Nexarian, line 56: So much nesting! Use breaks and continues to reduce nesting wherever possible. This is already done in my own branch, but it'll be awhile before that's merged.

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
...
+        # Build meter Lookup if available
+        meter_config = self.get_meters()
+        neurio = {}
+        c = 1000
+        # Loop through each Neurio device serial number
+        for n in lookup(self.status, ['neurio', 'readings']) or {}:
+            # Loop through each CT on the Neurio device
+            sn = n.get('serial', str(c))
+            cts = {}
+            c = c + 1
```
Nexarian, line 95: `c` can be computed using enumerate as well in the bigger, outer loop.

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
...
+        pvac_strings = self.__get_pvac_strings()
+
+        # Loop through each device serial number
+        for i, p in enumerate(lookup(self.status, ['esCan', 'bus', 'PVAC']) or {}):
+            if not p['packageSerialNumber']:
+                continue
+            packagePartNumber = p.get('packagePartNumber', str(i))
+            packageSerialNumber = p.get('packageSerialNumber', str(i))
+            pvac_name = f"PVAC--{packagePartNumber}--{packageSerialNumber}"
+            V_A = p['PVAC_Logging']['PVAC_PVMeasuredVoltage_A']
```
Nexarian, line 138: p['PVAC_Logging'] can be broken out into a temp variable for clearer readability.

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
...
+                "manufacturer": "TESLA",
+                "partNumber": packagePartNumber,
+                "serialNumber": packageSerialNumber,
+                "teslaEnergyEcuAttributes": {
+                    "ecuType": 296
+                }
+            }
+        return pvac
+
+    def get_pvs(self):
```
Nexarian, line 193: Why isn't this __?

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
...
+        pvs, _ = self.__get_pvs_and_pvac_strings()
+        return pvs
+
+    def __get_pvac_strings(self):
```
Nexarian, line 197: Rather than relying on magic knowledge of the tuple, return a plain old object that simply contains both?

pypowerwall/tedapi/vitals_dictionary.py
```diff
@@ -0,0 +1,517 @@
+import json
+from pypowerwall import __version__
```
Nexarian, line 2: Another general comment about this file: Overuse of the verb "get" -- Not all of these are actual "getters" (as in: They do no nor could we imagine a useful setter compliment).

Think about what the functions are doing. Are they `calculating`, `computing`, `deriving`, `aggregating`, `processing`, etc? Good function names usually have a "verb_dataStructure" theme, and `get` is vastly overused/abused!

Nexarian (changes requested):
Overall: I like the direction this is going.

Functionality wise, my biggest concern is the change in how the api lock works. That is a big shift and I'm not yet convinced we want that.

Otherwise: I encourage you to break this pull request up into smaller increments. You need not move every single function over to your more modularized version at once. Do one at a time. That makes it easier to review and makes the blast radius from any bugs much smaller.

pypowerwall/tedapi/__init__.py
mccahan: I'm leaning towards the trade-off that all the functions that were there right now are locking the whole thing, so it's probably better to make the decorator assume that and then if a function comes up later that only locks part that that would be the special case.

I think this is a bit more clear than having the extra `self_function` arg on every function, especially if you're ingesting it from the outside and using your IDE's function hinting.

pypowerwall/tedapi/decorators.py
mccahan: Not really, to get there you'd have to have ignored the convention of the rest of the file to be the only capitalized argument name, and you'd know it pretty quick

pypowerwall/tedapi/decorators.py
mccahan: That'd be my preference, but it scares people when I do that lol

pypowerwall/tedapi/__init__.py
Nexarian: The `self_function` thing was indeed awkward. It was my way to workaround the fact that there's no built-in self-referential variable for a function, but you're absolutely right that it reduces the readability and makes it harder to use.

How about a compromise: Add a warning to the comments of the decorator (and the others) that note that this affects the behavior of the entire function and that it doesn't support locking part of the function.

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
mccahan: It doesn't actually seem like it is, if you even so much as remove that space before "query" it fails. I think the big binary chunk on the next line is a checksum very specific to that query.

pypowerwall/tedapi/vitals_dictionary.py
mccahan: I'll probably leave this to you, then - I don't have this hardware, so I just lifted this from the existing file to move it here since I can't test it.

pypowerwall/tedapi/vitals_dictionary.py
mccahan: Ditto this one, without the hardware to test this I'm inclined to leave it as it was in the original file

pypowerwall/tedapi/vitals_dictionary.py
mccahan: Ditto, copied code I'm disinclined to touch

pypowerwall/tedapi/vitals_dictionary.py
mccahan: Easy enough!

pypowerwall/tedapi/vitals_dictionary.py
mccahan: I've never found myself confused by a get without a corresponding set (I only get annoyed when there's an unintended consequence that isn't indicated by the verb), but I don't care enough to die on that hill so I've just stripped "get" out of them entirely

Nexarian: It was something one of my mentors from Google taught me years ago. I've stuck to it since. I concede it is opinion.

pypowerwall/tedapi/vitals_dictionary.py
Nexarian: Sounds good. I'll come in later with my refactoring :)

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
Nexarian: Ah, I learned from https://github.com/jasonacox/pypowerwall/pull/152 that you can in fact change some things about the query without triggering a checksum invalidate, but sad that it can't be done here.

pypowerwall/tedapi/decorators.py
Nexarian: Use the right tool for the job. If some people protest that it's scary or they can't understand it; that's not YOUR problem! (IE: Healthy boundaries -- It's theirs!)

Split is a great way to do ersatz regex, and if this works it works, but any more complex than this and I'd probably say you'd want a regex.

pypowerwall/tedapi/decorators.py
Nexarian: Sounds good.

pypowerwall/tedapi/vitals_dictionary.py
Nexarian: Alright, I'll come in with my refactoring later on.

mccahan:
You're not wrong this should've been two PRs - I did pull the vitals part out into a separate branch just now (https://github.com/jasonacox/pypowerwall/compare/main...mccahan:chore/refactor-tedapi-vitals?expand=1) though I'm not sure I want to complicate a merge on this branch by removing/rebasing later. I did the side-by-side compare of `/vitals` on mine and they match 1:1, though I just have the single PW3 (no Neurio, etc).

The biggest risks (actual changes) are:
- Locking and the throttling
- Vitals tweaks for any hardware that doesn't match what I have

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
mccahan: It looks like in that one you're changing the _arguments_ to the query, not the query itself, e.g. here you can see it accepts an argument called `msaComp` of some `ComponentFilterSmsaSignals` type:

Screenshot 2025-03-23 at 15 18 56@2x

So you can change the parameters you're passing, but the query itself is checksummed.

pypowerwall/tedapi/__init__.py
mccahan: Fun story, I went to go compare some firmware stuff against the `main` branch and found that if I use `test_tedapi.py` and drop in a `gw.get_firmware_version()` throws an error because of the `self_function` argument.

Nexarian:
Re: Pain of rebasing. In the case of these gigantic refactors, the fact that you're scared to do it is a signal it's too big.

In those cases I often don't rebase, but re-create from whole cloth using the old branch as a reference, that's often easier.

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
Nexarian: I'm still surprised, given how locked down this generally is, that even that is allowed.

mccahan:
Nah it's not fear of a rebase, it's that the fun part is over and the ADHD will kick in and I'll lose interest if I'm playing rebase tag lol. We could PR in https://github.com/jasonacox/pypowerwall/compare/main...mccahan:chore/refactor-tedapi-vitals?expand=1 as the smallest digestible chunk, and deal with that merge conflict if that merges first, but otherwise I'll probably decline trying to split this up function-by-function. That feels like extra testing to me, dragging out a mystery piecemeal.

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
mccahan: Honestly I don't think it's a terrible approach for them, the alternatives are probably worse. Otherwise they'd have to be signing the arguments on the client side or building a more complicated API, and somebody's bound to break that anyway.

Nexarian:
@mccahan, this is really up to @jasonacox at this point :)

ADHD: https://www.youtube.com/watch?v=Zvqx9DfG9lU

jasonacox:
Sorry for the delay. Besides some personal emergencies, this is a lot to review. While I also like the work/direction, I have some concerns.

* From a code style perspecitive (yes, I'm guilty of not publishing anything to help) this is the first PR of the project using camel case for files. It feels off. Either we use it globally or we should reconsider the structure to match the existing.
* I've started worrying a bit more about the API locking mechanics. I think @Nexarian  mentions this and did a good job in a recently PR to help us improve, but I see things like https://github.com/jasonacox/pypowerwall/discussions/158 and wonder if we still have issues. Ccould be unrelated completely but still I'm concerned on how we test especialy in such a large PR. 
* Said by both of you, this is a big change. We have done them before based on great work by others to refactor, but I can't help but wonder how we could decouple some of this change to ensure time to bake between changes. How would we do that?

I would appreciate thoughts and perspectives. I'm still going through the code but I do have a lot going on now. Thanks for your patience.

pypowerwall/tedapi/ted_api_messages/BatteryComponentsMessage.py
Nexarian: I think the real reason is that this is utility-grade power equipment. It's a fact that's easy for us nerdy homeowners to forget, but they can't be letting every rando who thinks they have a "cool command" send arbitrary GraphQL to it. The attack surface area would be impossible to contain.

As much as I wish I could tweak the GraphQL to get parameters that I care about and/or remove ones I don't, I also understand why they do it.

Nexarian:
@mccahan Based on @jasonacox's comments: If your ADHD really does prevent you from splitting this up and staying the course, I'm happy to do it. I do this kind of thing all the time at my job, and I'm eventually going to have to do it with my refactor of the proxy.

But this is your change not mine, and I won't without your blessing.

mccahan:
Zero rush at all, hope things work out okay.

I swapped the camel casing (habit, didn't even notice), but otherwise if this is too much I'm happy to just abandon this PR. It's essentially all copying/pasting around, so beyond theoretical maintainability there's no gain. I'd personally err on the side of bundling it rather than stringing it across a few versions so it's easier to pinpoint when something happened if there's a bug report.

My test scenario has been running it side-by-side with the current production version and checking the results for each function; the substance parts of it would fail on requests because of changes to the caching, or the vitals wouldn't match up side-by-side (N.B. I have just the PW3 with none of the Neurio stuff).

Nexarian:
There is another way: Unit tests. Right now we have 0. We can cast the current behavior into unit tests, and if we have good enough coverage, refactor fearlessly :)

mccahan:
@Nexarian be my guest! I pulled a chunk out of it into #159, but if you'd like to split this one up you're more than welcome to. I wouldn't split it too far so you don't end up sprawling it over six versions trying to trace down any issues, but whatever works there. I'm also not at all hurt if we ditch this entirely.

(You just commented live lol) the only shortcoming to unit tests for this piece is that it wouldn't catch the primary risk of this particular PR, which is around testing it with hardware that I don't have. I think your hardware setup is more complicated, so maybe a few collections of responses from different systems to unit test vitals responses against could catch it, though that might become tedious quick if they start changing firmware responses.

Nexarian:
> I wouldn't split it too far so you don't end up sprawling it over six versions trying to trace down any issues
That's exactly what I would do. If you keep each change smaller and more manageable and isolated to unique minor versions, it's really easy to find where things went astray and identify the root cause and fix quickly. Version number increments are cheap, bugs are not.

Alright, I'll work on this then. I think merging in the spirit of your changes before my crazy proxy refactor would do all of us a lot of good.

jasonacox:
What should we do with this?

jasonacox:
The merge conflicts are significant. It would be a monumental task to reconcile at this point. Closing to start fresh.

PR #154, opened 2025-03-25
Prefix Animation Static Asset URLs for Proxy

mccahan:
Right now, the static assets for the animation (CSS/JS/images/fonts) are absolutely linked (e.g. `/app.js`) which poses some annoyance if you want to proxy the whole Grafana/animation kit out to the world from a single URL.

If we move them to a subdirectory (here I've used `/viz-static`), we can put a reverse proxy in front of the whole thing in one fell swoop and switch the backends we proxy to based on URL.

We can also set an API base URL to make things easier, so if we `export PROXY_API_BASE_URL=/pypowerwall/api`, one proxy pass is sufficient to catch the animation without having to define specific routes in your proxy config.

For example, using a Caddyfile and [Caddy](http://caddyserver.com/):
```
:8080 {
    handle_path /pypowerwall* {
        reverse_proxy localhost:8675
    }

    # Pass /api and /viz-static without stripping the paths, since the viz 
    reverse_proxy /api* localhost:8675
    reverse_proxy /viz-static* localhost:8675

    # All other requests go to localhost:9000
    reverse_proxy localhost:9000
}
```

With that, you can:
- http://localhost:8080/ goes to Grafana
- http://localhost:8080/pypowerwall goes to the animation, so you can `<iframe src="/pypowerwall">` from the Grafana dashboards

I did put the prefix in as a variable, but I didn't expose it as an environment variable because frankly I couldn't imagine someone needing to change it.


mccahan:
Actually I'm going to redo this one :D

PR #155, opened 2025-03-25
Add PROXY_BASE_URL option for reverse proxying

mccahan:
# The Problem
If you want to make everything nice and accessible remotely, right now you'd have to use two separate URLs, one for Grafana and one for pypowerwall to get the animation. If you try to put a reverse proxy in front of it, looking to get Grafana on `/` and pypowerwall on e.g. `/pypowerwall`, the animation won't function because all the URLs are hard-coded

# The (Proposed) Solution
- Move static assets into a subdirectory (this gives some flexibility with how you reverse proxy things, if you wanted to handle static assets separately it's a lot easier to differentiate them from Grafana's this way)
- Add a PROXY_BASE_URL environment variable
  - Use that in the URLs for the animation
  - Continue to support URLs directly to the base so that telegraf etc don't need to change
- Fix an annoying routing issue in the animation where it'll change the browser URL to the root

Now, you can use something like [this nginx container](https://github.com/mccahan/Powerwall-Dashboard/compare/feature/import-dashboards...mccahan:Powerwall-Dashboard:nginx) to make everything accessible on http://localhost:8090, where you can Cloudflare Tunnel or Ngrok or whatever to your heart's content. Iframe will be to `<iframe src="/pypowerwall">`

# Testing
Easiest way is probably:
- Check out the the Powerwall-Dashboard branch https://github.com/mccahan/Powerwall-Dashboard/tree/nginx
- Run through `./setup.sh`, hit enter a bunch to auto-import dashboards etc
- Go to `http://localhost:8090`
- `admin` / `admin`
- Observe:
  - The animation should load
  - You should start seeing data in Grafana


proxy/server.py
```diff
@@ -161,6 +162,11 @@
 logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
 log.setLevel(logging.INFO)
 
+# Ensure api_base_url ends with a /
+if not api_base_url.endswith('/'):
+    log.error("PROXY_BASE_URL must end with a '/'. Current value: '%s'" % api_base_url)
```
jasonacox, line 167: Why not just add the "/" to the end if it is missing? We can still log.info() the change.

mccahan: I 50/50'd on which way would be idiot-proofing, but your way is probably good; updated

proxy/server.py
```diff
@@ -279,15 +285,22 @@ def do_POST(self):
         global proxystats
         contenttype = 'application/json'
         message = '{"error": "Invalid Request"}'
-        if self.path.startswith('/control'):
+
+        # Get the request path and subtract the base URL. This is intentionally doing a replace on the URL we check
+        # instead of using the prefix inside of the checks to maintain the ability to use the regular URLs in addition to
+        # the proxied URLs, so Telegraf configs or existing integrations don't need to be modified.
+        request_path = self.path
+        request_path = request_path.replace(api_base_url, "/")
```
jasonacox, line 293: Something about this bothers me. The real intent is to replace just the prefix URI (e.g. `/pypowerwall/`) but the replace function will replace any occurence api_base_url within the string of self.path. Your example works because there aren't any URI's that use that, but if someone set this to `/system_status/` for example, it is going to wreck havoc with several APIs. I wonder if we should do something like this instead:
```python
        request_path = self.path
        if request_path.startswith(api_base_url):
            request_path = "/" + request_path[len(api_base_url):]
```

mccahan: Good thought, made that change in both places

jasonacox:
Thanks @mccahan  - I'll run some tests.

proxy/server.py
```diff
@@ -291,12 +291,12 @@ def do_POST(self):
...
-        # instead of using the prefix inside of the checks to maintain the ability to use the regular URLs in addition to
-        # the proxied URLs, so Telegraf configs or existing integrations don't need to be modified.
+        # If set, remove the api_base_url from the requested path. This allows installing the
+        # the proxy on a path without impacting the use of Telegraf or other integrations. Python 3.9+
         request_path = self.path
-        if request_path.startswith(api_base_url):
-            request_path = "/" + request_path[len(api_base_url):]
+        new_path = request_path.removeprefix(api_base_url)
+        if new_path is not request_path:
+            request_path = "/" + new_path
```
jasonacox, line 299: I was a little concerned about the string manupation required for every GET and POST. I ran some tests to ensure we aren't adding too much. 

I ran different load tests using basic install with animation rendering. I measured with:

```bash
export PID=1234 # process ID of python server.py
top -b -p $PID -n 60 -d 1 | grep $PID | awk '{sum+=$9} END {print "Average CPU: " sum/NR "%"}'
```

* Before this PR
CPU Usage - Average: 3.145%

*  After PR
CPU Usage - Average: 4.875%

Since we are pinned to Python 3.10, we can use a more performant string editor for prefix.  I made that update and it brings us back down to about where we were.

* With this PR after optimization
CPU Usage - Average: 3.2%

jasonacox:
Thanks for this addition @mccahan !

jasonacox:
jasonacox/pypowerwall:0.12.9t70

PR #156, opened 2025-03-25
Fix issue with visualization showing blank with multiple tabs

mccahan:
I have a problem with leaving way too many tabs open, so sometimes I'll end up with a few copies of the Grafana dashboard open with the corresponding pypowerwall dataviz. Sometimes you open one not realizing you've got 10 open already and it just shows a blank screen and you refresh and nothing and then you restart the browser and it's fine now, so... annoying.

The visualization app uses IndexDB for part of its old purpose (storing some kind of user data or something, things we don't use it for). As a result, it puts a lock in localstorage in IndexDB, which is shared at the URL level between tabs. If you have it open in too many tabs/iframes, sometimes it'll just go white and do nothing.

To reproduce:
- Open tabs to the visualization until you hit a white one
- Observe that even refreshing does nothing
- Close some of the older tabs until it suddenly starts working again
 
https://github.com/user-attachments/assets/ed6f484b-9ecd-4068-a360-3a0330700ac1

After fixing:
Go nuts, open as many tabs as you want.

https://github.com/user-attachments/assets/e3ba6fff-7394-4b76-9557-8f6b1972dd2d


Nexarian:
Seems to work for me... How did you ever find this? I've been hunting for this fix for a long time!

mccahan:
A couple hunches 😂. That it could be broken in a new tab in the same window but if I opened it in an Incognito window at the same time it would work fine without there being any change in the network traffic off the non-Incognito tab means it has to be client-side (also since you can't share local storage data between Incognito and normal environments). I checked for localstorage in the Application tab in devtools and saw nothing, searched the code for "lock" and saw nothing useful, so that left only IndexedDB as an option that could do a local lock.

jasonacox:
Absolutely brilliant!!!  ❤️ 🙏

jasonacox:
The merge conflict is due to https://github.com/jasonacox/pypowerwall/pull/155 and the vendor.js moving into a new path. @mccahan can you resolve conflict so we can merge?

mccahan:
@jasonacox good to go!

jasonacox:
jasonacox/pypowerwall:0.12.9t69-beta

jasonacox:
jasonacox/pypowerwall:0.12.9t70

PR #159, opened 2025-03-31
Refactor TEDAPI Vitals for Readability

mccahan:
This is the most-easily-separable part of #153 - just pulling the vitals function out into a separate file.

**Testing**
- Run this branch alongside an existing version
- Open `/vitals` in each, and compare side-by-side
  - I did this using a diff tool, and noted that the only difference between the two would be small measurement differences based on the timing of the data pull

N.B. I only have a PW3 to test on, none of the Neurio stuff, so I can't test those parts (though they should be essentially unchanged)


jasonacox:
We need to move the METER_Z (TEMSA block) code into vitals.py.

jasonacox:
The merge conflicts here are too great at this point. I appreciate the attempt @Nexarian but I'm closing this to start fresh.

PR #161, opened 2025-04-07
Add fan speed routes and update proxy version to t71

jasonacox:
### Proxy t71 (6 Apr 2025)

* Add routes for fan speeds: `/fans` and `/fans/pw` (simple enumerated values for dashboard)


jasonacox:
jasonacox/pypowerwall:0.12.9t71

PR #164, opened 2025-04-15
Fix kwargs bug

F1p:
Fix for 
TypeError: PyPowerwallTEDAPI.vitals() got an unexpected keyword argument 'force'


jasonacox:
Thanks @F1p! Great fix.

PR #166, opened 2025-04-16
Issue #162: add /pw/XXX endpoints to expose Powerwall() API methods

JohnJ9ml:
Issue #162: add /pw/XXX endpoints to expose Powerwall() API methods as JSON-returning endpoints

### DIFF
--- a/proxy/server.py
+++ b/proxy/server.py
@@ -68,6 +68,33 @@ ALLOWLIST = [
 DISABLED = [
     '/api/customer/registration',
 ]
+PW_ALLOWLIST = [
+    '/pw/level',
+    '/pw/power',
+    '/pw/site',
+    '/pw/solar',
+    '/pw/battery',
+    '/pw/battery_blocks',
+    '/pw/load',
+    '/pw/grid',
+    '/pw/home',
+    '/pw/vitals',
+    '/pw/aggregates',
+    '/pw/temps',
+    '/pw/strings',
+    '/pw/din',
+    '/pw/uptime',
+    '/pw/version',
+    '/pw/status',
+    '/pw/system_status',
+    '/pw/grid_status',
+    '/pw/site_name',
+    '/pw/alerts',
+    '/pw/is_connected',
+    '/pw/get_reserve',
+    '/pw/get_mode',
+    '/pw/get_time_remaining',
+]
 web_root = os.path.join(os.path.dirname(__file__), "web")
 
 # Configuration for Proxy - Check for environmental variables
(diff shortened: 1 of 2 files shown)


JohnJ9ml:
My first attempt at any of this. consider it a learning exercise.

jasonacox:
Great job @JohnJ9ml !  I have some suggestions but basics are here.

proxy/server.py
```diff
@@ -730,6 +757,81 @@ def do_GET(self):
                 message = json.dumps(fans)
             else:
                 message = '{}'
+        elif self.path in PW_ALLOWLIST:
```
jasonacox, line 760: Actually, we dont need this we can use:

```python
elif self.path.startswith('/pw/'):
```

We can then eliminate the big array and I'm thinking about just using a lambda function map. I've been planning to do that for all the rest, but this would be a good start.

jasonacox: I'll submit a commit with those adjustments.

JohnJ9ml: I knew this check would suffice, but I didn't want to break consistency - if there was already one array, it would be a shame to impose some other style on something I was just enhancing.  I anticipated there was something more table-driven in mind.

Thanks for making the adjustments - you figured that would save me weeks of trying to figure out how to do it myself, right?

jasonacox: Actually, I just had Copilot rewrite it. ;)

JohnJ9ml:
I"ve been using a local copy of this for a week now, and it works well for me. The one anomaly I've noted is that occasionally the HTTP connection kicks up errors, and I end up getting empty data back. My suspicion is that this little box sitting right next to the Dashboard host combined is just a bit too much for the inverter to handle. I look forward to the next Dashboard update that includes this fix, at which time I'll shut my proxy off and see how everything handles the load of my process running against the same proxy that's supporting the Dashboard.

Odd note: when the HTTP connection seems to have woes, it's generally (but not always) at night - i.e. when the sun is down and one would think there's seemingly not as much data for the inverter to process. Amusing, but seems a general pattern at present. (Then again, the next 23 coin flips could all come up tails, at which point.....)

jasonacox:
Hi @JohnJ9ml  - What error do you see in the logs when this happens?

JohnJ9ml:
The errors are generally timeouts and no route to host errors.

I'm just accepting right now that two proxies together are likely abusing the poor little inverter antenna, and waiting out the next release. 

To its credit, the proxy hangs out, waits its time, retries for the next request, and eventually succeeds. I'm pleased it stays on its feet. I have to do a bit of filtering on my end because the data comes back with 0 values (for the power() call, for example), but that's perfectly reasonable in my mind for a system meant to run for months at a time.

PR #181, opened 2025-05-26
Initial simple unit test

Nexarian:
The best unit tests are motivated by bugs you've found in working code. You find the bug, then fix it. In this example, the "no location" branch of the `current_power` function never worked. It always outputted an empty dictionary (wrong!). So let's both fix that BUG and patch the hole with a unit test for all time.

Additionally, the intent of this is to get the ball rolling on a more comprehensive unit test suite for pypowerwall in general to prevent regressions, especially now that there are so many different device configurations, it's unreasonable to expect an intrepid dev who is refactoring something to test every configuration known for Tesla devices!

### DIFF
--- a/pypowerwall/tedapi/__init__.py
+++ b/pypowerwall/tedapi/__init__.py
@@ -52,7 +52,7 @@ import threading
 import time
 from functools import wraps
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


Nexarian:
Got the test runner working. Now ready for review.

PR #193, opened 2025-07-11
Show an image of a Powerwall 3 instead of a Powerwall 2 if it is a PW3.

JEMcats:
In the webpage for the proxy, there is a live power flow diagram with an image of a Powerwall. This defaults to a Powerwall 2, I have updated it to check the ```/stats``` for the ```"pw3"``` value and if it is true change the photo to a Powerwall 3.


jasonacox:
Nice! Let's get this queued up for testing and the next release. Thanks!

jasonacox:
jasonacox/pypowerwall:0.13.2t78

PR #213, opened 2025-09-09
Pytest now runs all tests. Also add line and branch coverage

Nexarian:
# Summary
Before this PR, not all tests ran. Pytest wasn't finding them due to an over-specified test path search.

Additionally, we weren't measuring line and branch coverage. This fixes that. Upload to codecov as an easy way to see the coverage results in a browser. This should be free for public repos.

Finally, upgrade github actions plugins and make minor improvements to the test.yml workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,22 +5,34 @@ on:
   pull_request:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   tests:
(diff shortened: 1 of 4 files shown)


Nexarian:
@jasonacox Penny for your thoughts on this. Plan is to improve unit test coverage on do_GET in the proxy. Then, after that's done, I will refactor it to add my multi-inverter code as a way to ensure we're not breaking things.

jasonacox:
Thanks @Nexarian ! Good idea. Merged. I added CODECOV_TOKEN for reports.

https://app.codecov.io/gh/jasonacox/pypowerwall?search=&displayType=list

image

PR #216, opened 2025-09-15
Add initial unit tests to the proxy

Nexarian:
# Summary

Initial unit test, as simple as I could get it. Unfortunately, it does require some modifications to the proxy to get working.

First unit test chosen because it's easy and closed-scope: The stats endpoints.

Biggest problem: The code in the proxy is no longer fully compatible with Python 3.8: https://indhumathychelliah.com/2020/10/12/new-string-methods-to-remove-prefixes-and-suffixes-in-python-3-9/

The `do_GET` handler uses `removeprefix`. This was already merged, so the proxy has been broken on 3.8 for a while. This tells me that we can justifiably drop it. Clearly nobody was actually using this with Python 3.8 any longer, otherwise we'd have gotten bug reports.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
 
     steps:
       - uses: "actions/checkout@v4"
--- a/proxy/server.py
+++ b/proxy/server.py
@@ -104,10 +104,10 @@ import urllib3
 # 3. Executing from within the proxy directory (plain module import)
 try:  # Prefer relative when executed as a package module
     from .transform import get_static, inject_js  # type: ignore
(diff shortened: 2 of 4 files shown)


Nexarian:
@jasonacox This is ready for you to take a look at.

Nexarian:
@jasonacox Are you the only one who can call for the github copilot review? I looked around to try to find the command/UI option and I didn't see it.

jasonacox:
> @jasonacox Are you the only one who can call for the github copilot review? I looked around to try to find the command/UI option and I didn't see it.

Yes, unfortunately. I do like to run it as it has found some non-obvious bugs. But also gets things worngs too. :) 

I made a minor change to this to accomodate my local test harness for the proxy. I suspect others may use it too (to run local vs via container).

jasonacox:
Thanks for this @Nexarian !

Nexarian:
You're welcome! As I've mentioned before, the long-term plan here is to have all of "do_GET" unit tested... So that I can then refactor it. Which is a prelude to adding the functionality for multiple inverters/gateways/etc.

PR #218, opened 2025-09-23
Continue proxy unit testing tranche 2

Nexarian:
Merge https://github.com/jasonacox/pypowerwall/pull/216 first


Nexarian:
Ok this works now, but I'll split it up into more sensical pieces. I fell into a rabbit hole on the simulator and making sure that even forks use the proper image from the repo root.

PR #223, opened 2025-10-05
Add curl to Dockerfiles to use in healthchecks instead of wget

goodoldme:
Added curl to Dockerfiles to use in healthchecks instead of wget to fix [issue ](https://github.com/jasonacox/Powerwall-Dashboard/issues/693)


jasonacox:
jasonacox/pypowerwall:0.14.2t83

PR #243, opened 2025-12-09
Update all Python versions to be consistent

Nexarian:
- Update versions of github actions plugins
- Remove 3.8, add 3.13


jasonacox:
> Do we want linting to be more selectively run for some reason?

No. Intent is for lint for everything. Thanks for the fix!

PR #275, opened 2026-04-12
v0.15.1 - Code Quality and Build Pipeline Improvements

jasonacox:
## v0.15.1 - Code Quality and Build Pipeline Improvements

* Fix: Remove duplicate stub methods `get_grid_charging()` and `get_grid_export()` in `pypowerwall_tedapi.py` that were left over from a merge — the real implementations (reading/writing config via v1r transport) were already present and being shadowed
* Fix: Update `pwsimulator` `stub.py` to use `ssl.SSLContext` API replacing the removed `ssl.wrap_socket()` call, which caused the simulator container to silently exit on Python 3.12+
* Fix: Remove `linux/arm/v7` platform from `pwsimulator` Docker build (`upload.sh`) — platform is no longer supported
* Fix: Correct protobuf runtime dependency — `protobuf>=3.20.0` was misleading; pb2 files generated with 4.25.x require `protobuf>=4.25.1,<5`
* Add: `.pylintrc` with `[MESSAGES CONTROL]` disable list (restored), `[SIMILARITIES]` config, and `ignore-paths` to skip auto-generated `*_pb2.py` files
* Add: `tools/gen_proto.sh` — script to regenerate all `*_pb2.py` files from `.proto` sources using pinned `grpcio-tools`
* Add: `tools/requirements-tools.txt` — pinned dev tools (`grpcio-tools<1.64`, `protobuf<5`) to ensure pb2 files are generated consistently with a compatible protobuf version
* Add: `.pre-commit-config.yaml` — pre-commit hooks for protobuf regeneration and `pylint -E` checks on `pypowerwall/` and `proxy/` before every commit
* Add: `.github/workflows/check-protobuf.yml` — CI workflow to verify committed `*_pb2.py` files are in sync with their `.proto` sources


Nexarian:
@jasonacox One quick thought: Should we have the protobuf build just be automatic as part of the Dockerfile build? That will guarantee that no version incompatibilities arise when it's actually loaded.

jasonacox:
View at:
https://pypi.org/project/pypowerwall/0.15.1/

jasonacox/pypowerwall:0.15.1t88

jasonacox:
Hi @Nexarian  - Actually, the Dockerfile is essentially doing a `pip install pypowerwall==0.15.1` so the pb2 files will be coming from the library. This update adds a pre-commit build and a GH action protobuf test. 

However, I did spot a gotcha in setup.py that makes it over restrictive (`protobuf>=4.25.1,<5`).  I'll need to fix that.

PR #276, opened 2026-04-12
v0.15.2 - Protobuf Support

jasonacox:
## v0.15.2 - Minor Fixes

* Fix: Remove `<5` upper cap on `protobuf` runtime dependency — constraint is now `protobuf>=4.25.1`; pb2 files generated with 4.25.x are compatible with 5.x, 6.x, and 7.x runtimes (confirmed tested up to 7.34.1) and the cap was causing pip conflicts for users with newer protobuf versions installed (e.g. via TensorFlow)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,14 +2,14 @@
 
 ## v0.15.2 - Minor Fixes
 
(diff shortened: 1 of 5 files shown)


jasonacox:
View at:
https://pypi.org/project/pypowerwall/0.15.2/

jasonacox/pypowerwall:0.15.2t88

PR #277, opened 2026-04-12
feat: add go_off_grid() and reconnect_grid() for Powerwall island mode control

bolagnaise:
## Summary

Adds `go_off_grid()` and `reconnect_grid()` methods to the `Powerwall` class for physical grid contactor control.

## Discovery

Reverse-engineered via mitmproxy capture of the Tesla mobile app and confirmed on live systems:

- `setIslandModeRequest(mode=6, force=True)` — physically opens the grid contactor (off-grid)
- `setIslandModeRequest(mode=1, force=False)` — physically closes the grid contactor (reconnect)
- `mode=2` (previously assumed by the community to be off-grid) is accepted by the gateway but **does not physically operate the contactor** on either PW2 or PW3
- `force=True` is required for off-grid — without it the gateway acknowledges the command but does not open the contactor

## Signed Commands Required (Both PW2 and PW3)

On both Powerwall 2 and Powerwall 3, the command must be delivered as a **signed RoutableMessage** via the cloud `device_command` endpoint:

1. Pair an RSA-4096 key with the gateway via `add_authorized_client_request` (requires physical DC isolator toggle for verification — key must reach state=3)
2. Build a `MessageEnvelope` with `deliveryChannel=2` (HERMES_COMMAND), `sender.authorizedClient=1`, `recipient.din=<gateway_din>`, `teg.setIslandModeRequest(mode=6, force=True)`
3. Wrap in a `RoutableMessage` (field 6=domain 7, field 10=envelope bytes) with RSA-PKCS1v15-SHA512 signature over TLV-encoded payload
4. Base64-encode and POST as the `routable_message` field (not `energy_device_message`) in `POST /api/1/energy_sites/{id}/device_command`

### What does NOT work
- **Unsigned `grpc_command` via `/command`** — gateway returns "signature required: no authorized client public key"
- **Local REST `/api/v2/islanding/mode`** — requires installer-level auth, customer role gets 403
- **`triggerIslandingBlackStartRequest`** — blocked by Tesla's powergate (400) on all cloud paths
- **`mode=2`** — sets a preference only, does not operate contactor
- **`force=False`** — gateway acknowledges but does not island
- **Cloud backup+100% reserve** — does not physically island

## This PR

Adds the public API surface (`go_off_grid()`, `reconnect_grid()`) with `hasattr` delegation to the backend client. Backend implementations (fleetapi, tedapi, cloud, local) are left as stubs — this PR documents the protocol so implementers know exactly what to build.

## Review Fixes

- Updated docstrings: PW2 uses the same signed cloud path as PW3 (not local REST)
- Added `force=True` documentation
- Error messages now include backend class name per Sam's review

## Confirmed Working

- **Powerwall 2** firmware 26.10.0 — signed `routable_message` via `device_command`, mode=6, force=True
- **Powerwall 3** firmware 26.2.1 — same method

Both confirmed via [PowerSync](https://github.com/bolagnaise/PowerSync).

### DIFF
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -967,7 +967,7 @@ class Powerwall(object):
         """
         return self.client.get_grid_export()
 
-    def go_off_grid(self) -> Optional[dict]:
+    def go_off_grid(self, confirm: bool = False) -> Optional[dict]:
         """
         Physically disconnect the Powerwall from the grid (open contactor).
 
@@ -977,6 +977,9 @@ class Powerwall(object):
         dropout during the contactor transition.
 
         Note:
+            This operation requires explicit confirmation. Pass
+            confirm=True to send the command.
+
             On both Powerwall 2 and Powerwall 3, this requires a signed
             RoutableMessage via the cloud device_command endpoint's
             routable_message field. The unsigned grpc_command path is
@@ -991,6 +994,9 @@ class Powerwall(object):
         Returns:
             Dictionary with operation results, or None if unsupported.
         """
+        if not confirm:
+            log.error("go_off_grid requires confirm=True")
+            return None
         if hasattr(self.client, 'go_off_grid'):
             return self.client.go_off_grid()
         log.error("go_off_grid is not supported by %s backend", type(self.client).__name__)
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -967,6 +967,47 @@ class Powerwall(object):
         """
         return self.client.get_grid_export()
 
+    def go_off_grid(self) -> Optional[dict]:
+        """
+        Physically disconnect the Powerwall from the grid (open contactor).
+
+        Sends setIslandModeRequest(mode=6) which physically opens the grid
+        contactor, islanding the home. Solar continues producing and the
+        battery serves home load. Causes ~30s solar dropout during the
+        contactor transition.
+
+        Note:
+            On Powerwall 3, this requires a signed RoutableMessage via the
+            device_command endpoint's routable_message field. The unsigned
+            grpc_command path is accepted but does not physically operate
+            the contactor.
+
+            On Powerwall 2, this uses the local REST endpoint
+            /api/v2/islanding/mode.
+
+        Returns:
+            Dictionary with operation results, or None if unsupported.
+        """
+        if hasattr(self.client, 'go_off_grid'):
+            return self.client.go_off_grid()
+        log.error("go_off_grid is not supported in the current mode")
+        return None
+
+    def reconnect_grid(self) -> Optional[dict]:
+        """
+        Reconnect the Powerwall to the grid (close contactor).
+
+        Sends setIslandModeRequest(mode=1) which physically closes the
+        grid contactor, reconnecting the home to the utility grid.
+
+        Returns:
+            Dictionary with operation results, or None if unsupported.
+        """
+        if hasattr(self.client, 'reconnect_grid'):
+            return self.client.reconnect_grid()
+        log.error("reconnect_grid is not supported in the current mode")
+        return None
+
     def _validate_init_configuration(self):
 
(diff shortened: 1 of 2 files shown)


jasonacox-sam (changes requested):
The PR description is excellent — the mitmproxy-captured protocol notes, the mode=2 correction, and the signed RoutableMessage requirements for PW3 are genuinely valuable documentation. That alone is a contribution worth merging.

Two things to address before this lands:

**1. Docstring contradicts PR body on PW2**

The `go_off_grid()` docstring says:
> On Powerwall 2, this uses the local REST endpoint `/api/v2/islanding/mode`.

But the PR body explicitly says that endpoint requires installer-level auth (`Home_Owner` gets `403`) and that the working PW2 path is the Tesla cloud API (`backup_reserve_percent: 100` + `default_real_mode: backup`). The docstring should reflect what actually works — otherwise implementers will waste time on the 403 path.

**2. Safety guardrail — consider a `confirm` parameter**

`go_off_grid()` physically opens a grid contactor. A stray call (mis-typed variable, accidental loop iteration, testing against prod) causes a real outage — ~30s solar dropout plus whatever happens downstream. Other pypowerwall methods that set reserve percentage or operating mode don't have this kind of physical consequence.

Worth considering: `go_off_grid(confirm=False)` that raises or returns `None` unless `confirm=True` is explicitly passed. Makes the intent legible at the call site and prevents accidents. Not a blocker if you disagree, but worth a decision.

**Everything else looks good.** The `hasattr` delegation pattern is consistent with the rest of the codebase, `log.error` + `return None` on unsupported is the right fallback, and stub-first is a reasonable way to land the public API surface before the backend implementations arrive.

— @jasonacox-sam

jasonacox-sam:
Chiming in on Copilot's two comments (lines 994 and 1009) — they're making the same valid point on both methods: the unsupported-path log message doesn't tell you *which* backend is active, which makes user reports hard to diagnose.

I'd take the first half of the suggestion and skip the second. Including the client type in the message is genuinely useful:

```python
def go_off_grid(self) -> Optional[dict]:
    if hasattr(self.client, 'go_off_grid'):
        return self.client.go_off_grid()
    log.error("go_off_grid is not supported by %s backend", type(self.client).__name__)
    return None

def reconnect_grid(self) -> Optional[dict]:
    if hasattr(self.client, 'reconnect_grid'):
        return self.client.reconnect_grid()
    log.error("reconnect_grid is not supported by %s backend", type(self.client).__name__)
    return None
```

Where I differ from Copilot: keep `log.error`, not a softer level. Calling a method that physically operates a grid contactor on a backend that doesn't support it is an error condition — the stakes are different from a read-only method quietly returning `None`. The user needs to know something went wrong.

— Sam

bolagnaise:
Gimme a Few hours @jasonacox, i've got access to a PW2 so will confirm the correct method for PW2 owners, its the same process as PW3, local REST is dead. Once i confirm i will update the PR. PW3 is confirmed working though!

bolagnaise:
Both off-grid and reconnect confirmed working on PW2 and PW3. Full round-trip tested multiple times. First open-source implementation of remote PW islanding control!

**Implementation note:** This requires the `device_command` endpoint with the `routable_message` field — the unsigned `grpc_command` path via `/command` reaches the gateway but is rejected with "signature required". I don't see `device_command` support in the fleetapi client currently (only `/command` in `v1r_register.py` for pairing). The
signing logic exists in `tools/tedapi-lan/` but for local delivery only. A cloud `device_command` path with signed
`routable_message` would need to be added to the fleet API client for this to work end-to-end.

   Changes per review:
   - Docstrings corrected — PW2 uses the same signed cloud path as PW3 (local REST requires installer auth)
   - Error messages include backend class name per @jasonacox-sam's suggestion
   - PR body updated with complete findings including what doesn't work

jasonacox:
Thanks for the update @bolagnaise - holding or next update release.

PR #278, opened 2026-04-16
Fix TypeError when solar signal value is None on PW3 without solar panels

anopheles:
On Powerwall 3 systems without solar panels, PCH_PvVoltage and PCH_PvCurrent signals have a None value. The comparison `signal['value'] > 0` throws TypeError: '>' not supported between instances of 'NoneType' and 'int', causing get_api_meters_aggregates() to fail and return null data.

Fix by guarding with `is not None` before the comparison in all three affected files.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,7 +1,23 @@
 # RELEASE NOTES
 
+## v0.15.3 - PW3 No-Solar None Handling
+
+* Update scanner to use cidr by @Nexarian in https://github.com/jasonacox/pypowerwall/pull/266
+* Add go_off_grid() and reconnect_grid() for Powerwall island mode control by @bolagnaise in https://github.com/jasonacox/pypowerwall/pull/277
+* Fix: Prevent `TypeError` in Powerwall 3 TEDAPI vitals parsing when `PCH_PvVoltage*` or `PCH_PvCurrent*` signals are `None` on systems without solar panels - by @anopheles in https://github.com/jasonacox/pypowerwall/pull/278
+     * `get_pw3_vitals()` now guards `None` values before performing `> 0` comparisons
+     * PV measured voltage, current, and power now safely report `0` when the gateway returns missing PV values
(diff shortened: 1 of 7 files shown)


jasonacox-sam:
Thanks for this fix, @anopheles! 🙏 Clean and targeted — exactly the right approach for this edge case on PW3 systems without solar panels.

One thing worth noting from the Copilot review: it flagged that the existing `get_pw3_vitals()` tests under `pypowerwall/tests/tedapi/` only cover numeric PV voltage/current values. It suggests adding a regression test case where `PCH_PvVoltage*` and/or `PCH_PvCurrent*` have `value: None` — confirming the path no longer raises and that PV measured values/power are reported as `0`. That would be a solid addition to lock in this fix long-term.

If you can add that test, this looks good to merge after a practical test on a PW3 system without solar. Thanks again for contributing to the community — this kind of defensive fix helps a lot of PW3 users who run into this silently! ⚡

anopheles:
> Thanks for this fix, @anopheles! 🙏 Clean and targeted — exactly the right approach for this edge case on PW3 systems without solar panels.
> 
> One thing worth noting from the Copilot review: it flagged that the existing `get_pw3_vitals()` tests under `pypowerwall/tests/tedapi/` only cover numeric PV voltage/current values. It suggests adding a regression test case where `PCH_PvVoltage*` and/or `PCH_PvCurrent*` have `value: None` — confirming the path no longer raises and that PV measured values/power are reported as `0`. That would be a solid addition to lock in this fix long-term.
> 
> If you can add that test, this looks good to merge after a practical test on a PW3 system without solar. Thanks again for contributing to the community — this kind of defensive fix helps a lot of PW3 users who run into this silently! ⚡

Added a test under `pypowerwall/tests/tedapi/` that validates my changes. This was tested with a PW3 system.

jasonacox:
Thanks for this @anopheles !

View at:
https://pypi.org/project/pypowerwall/0.15.3/

Proxy:
jasonacox/pypowerwall:0.15.3t88

PR #287, opened 2026-05-07
docs: note DHCP server requirement for Option 5 wired LAN

### LINKED ISSUE #286: Add: Under Option 5 in the README - DHCP server is required
Okay, now I've got it. I think the really important piece of information missing under Option 5 in the README is that a DHCP server is required, and that the PW does not have a static IP address within that range.

Thanks again!

_Originally posted by @TivaliEU in https://github.com/jasonacox/pypowerwall/discussions/280#discussioncomment-16836345_

jasonacox-sam:
The Powerwall 3 does **not** have a static IP address on the vendor subnet — it relies on DHCP to obtain one. Without a DHCP server running on the vendor subnet, the wired LAN endpoints are unreachable.

## Changes
- Added a prominent ⚠️ DHCP note to the **Network Requirements (Wired LAN)** section under Option 5
- Clarifies that a DHCP server (e.g., `dnsmasq` or a Linux NAT router) must hand out addresses in the `10.42.1.x` or `10.45.1.x` range

Closes #286

Reported by @TivaliEU in discussion #280.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ Option 5 has two sub-modes:
 
 The Powerwall 3 gateway has a wired Ethernet port on the TEG (Tesla Energy Gateway) unit that operates on an internal vendor subnet — typically `10.42.1.0/24` or `10.45.1.0/24`. This is **not** your home LAN IP.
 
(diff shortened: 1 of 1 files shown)


jasonacox:
Thanks, Sam! 😉

PR #291, opened 2026-05-10
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `tools/fleetapi/index.html:L24`

In tools/fleetapi/index.html, the 'code' parameter from the URL is directly inserted into the HTML without sanitization. An attacker could craft a malicious URL with JavaScript code in the 'code' parameter to execute XSS attacks.

## Solution

Sanitize the code parameter before inserting it into HTML. Use textContent instead of document.write() or properly escape the value with a library like DOMPurify.

## Changes

- `tools/fleetapi/index.html` (modified)
- `proxy/transform.py` (modified)

### DIFF
--- a/proxy/transform.py
+++ b/proxy/transform.py
@@ -20,6 +20,14 @@ def get_static(web_root, fpath):
     if fpath.startswith("/"):
         fpath = fpath[1:]
     freq = os.path.join(web_root, fpath)
+
+    # Prevent path traversal attacks by validating resolved path stays within web_root
+    real_path = os.path.realpath(freq)
+    real_web_root = os.path.realpath(web_root)
+    if not real_path.startswith(real_web_root + os.sep) and real_path != real_web_root:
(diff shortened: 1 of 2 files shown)


jasonacox-sam (approved):
Thanks for this fix! Clean and well-targeted. Here's what I see:

- **XSS via `document.write()` on `page` param** — replaced with safe DOM insertion. Correct fix, eliminates the injection vector entirely.
- **XSS via `document.write()` on `style` param in proxy handlers** — same pattern, same correct approach.

CI is green ✅. Nice work.

**Optional follow-up:** There are a handful of other `document.write()` calls in `index.html` that aren't currently reachable from user input but could become risky if the code evolves. Worth auditing when convenient, but not blocking.

— Sam ⚡

PR #301, opened 2026-05-13
fix: v1r Owner API login — use tesla_auth WebView instead of teslapy browser redirect

### LINKED ISSUE #300: v1r Owner API registration fails — tesla:// redirect not captured by browser
## Bug Report

**Reported in:** #299 (thanks @marcg1)

### Problem

When running `python -m pypowerwall setup -v1r` and selecting Owner API authentication, the user is asked to open a Tesla login URL in their browser. After successful login, Tesla redirects to `tesla://auth/callback?code=...` — but browsers cannot handle the `tesla://` custom URL scheme. The browser either:

- Refreshes and stays on the same page (showing no code parameter in the URL)
- Shows an error about the unrecognized URL scheme

The user then pastes back the displayed URL (which has no `code` parameter) and gets:
```
Login failed - (missing_code) Missing code parameter in response
```

### Root Cause

`v1r_register.py` → `owner_api_login()` uses `teslapy.Tesla` to generate the OAuth URL. teslapy uses `tesla://auth/callback` as the redirect URI — this only works when the redirect is intercepted by a native WebView (as `tesla_auth.py` does for cloud mode setup). It does NOT work in a regular web browser.

### Fix

The Owner API path in `v1r_register.py` should use `tesla_auth.py`'s login flow (which uses a native WebView with navigation interception on macOS, or pywebview on Linux/Windows) instead of `teslapy`'s browser-based approach.

This is the same fix that was applied to cloud mode setup — the OAuth flow changed and we need to intercept the `tesla://` callback via a WebView, not rely on the browser to handle it.

### Steps to Reproduce

1. Run `python -m pypowerwall setup -v1r`
2. Select Owner API (option 1)
3. Open the generated URL in a browser
4. Log in to Tesla
5. Browser refreshes but URL shows no `code` parameter
6. Paste URL back → "Missing code parameter in response"

### Environment

- Powerwall 3 (confirmed by reporter)
- Affects all platforms where Owner API is used for v1r registration
- The Fleet API path (option 2) may also be affected if it uses a browser redirect

— Sam 🌊

jasonacox-sam:
## Summary

Fixes #300 (reported in discussion #299)

The v1r Owner API registration path (`python -m pypowerwall setup -v1r` → option 1) fails because it uses `teslapy`'s browser-based OAuth flow. This generates a login URL that redirects to `tesla://auth/callback?code=...` after authentication — but browsers can't handle the `tesla://` custom URL scheme. The user sees the page refresh with no `code` parameter and gets:

```
Login failed - (missing_code) Missing code parameter in response
```

## Root Cause

`v1r_register.py` → `owner_api_login()` called `teslapy.Tesla` to generate the OAuth URL and expected the user to paste back the redirect URL from the browser. The `tesla://` scheme only works when intercepted by a native WebView.

## Fix

Switch `owner_api_login()` to use `tesla_auth.py`'s native WebView PKCE flow — the **same approach already working for cloud mode setup** (`python -m pypowerwall setup`). The WebView intercepts the `tesla://` callback internally, so the browser redirect issue doesn't apply.

Changes:
- Replaced `teslapy.Tesla` browser-based flow with `tesla_auth.login()` WebView flow
- Added cached token reuse (reads `.pypowerwall.auth` directly, refreshes if expired)
- Falls back to headless/paste mode for remote sessions (same as cloud setup)
- No changes to the Fleet API path (option 2)

## Testing

- Syntax check passes
- The auth flow now matches cloud mode setup which is confirmed working
- Remote/headless sessions still supported via `tesla_auth.login()` auto-detection

— Sam 🌊

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,11 @@
 # RELEASE NOTES
 
+## v0.15.7 - v1r Owner API Login Fix
+
+* Fix: v1r Owner API registration (`python -m pypowerwall setup -v1r` → option 1) now uses the native `tesla_auth` WebView PKCE flow instead of the broken `teslapy` browser redirect. The `tesla://` custom URL scheme callback is intercepted by the WebView, eliminating the "missing_code" login failure (#300, reported in discussion #299)
+* Fix: Cached token lookup in `owner_api_login()` now selects the account matching the requested `email` argument instead of always using the first entry in `.pypowerwall.auth`
+* Bump library version to `0.15.7`
+
 ## v0.15.6 - Reserve Percent Scaling Fix + CLI Redesign
 
 * Fix: `set_operation()` reserve percent scaling — reverse Tesla App scaling (0–100%) to raw API scale (5–100%) only in TEDAPI v1r mode, avoiding incorrect round-trip values in cloud and FleetAPI modes
--- a/pypowerwall/__init__.py
+++ b/pypowerwall/__init__.py
@@ -89,7 +89,7 @@ import time
 from json import JSONDecodeError
 from typing import Optional, Union
 
-version_tuple = (0, 15, 6)
+version_tuple = (0, 15, 7)
 version = __version__ = '%d.%d.%d' % version_tuple
 __author__ = 'jasonacox'
 
--- a/pypowerwall/v1r_register.py
+++ b/pypowerwall/v1r_register.py
@@ -498,11 +498,11 @@ def step4_register_key(token, energy_site_id, public_key_der, fleet_api_base):
 
 def owner_api_login(email=None, authpath="", force_reauth=False):
     """
-    Authenticate with the Tesla Owner API using teslapy.
+    Authenticate with the Tesla Owner API using tesla_auth.
 
-    This is the same login flow as 'python -m pypowerwall setup' (Cloud Mode).
-    No developer app or hosted key is required — just your Tesla account
-    email and password.
+    Uses the same native WebView PKCE flow as cloud mode setup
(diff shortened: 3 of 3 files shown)


jasonacox:
@jasonacox-sam Please review copilot comments and address as needed. Also, please increment the library version number and add RELEASE.md section to prep for release. I will begin testing and if they pass, we can merge.

jasonacox-sam:
Pushed commit `47c1035` with the following updates:

**Copilot review — addressed:** The cache lookup in `owner_api_login()` now selects the account matching the `email` argument when provided, falling back to the first cached account only when no email is specified. Valid catch — applied.

**Version bump:** `0.15.6` → `0.15.7` in `pypowerwall/__init__.py`

**RELEASE.md:** Added `v0.15.7` section covering the auth flow fix and the email-selection fix.

Ready for testing. 🚀

— Sam ⚡

jasonacox-sam:
Valid point on the cache lookup — currently it grabs the first entry regardless of email. In practice, pypowerwall is almost always single-account, but adding email-based selection is more correct. I'll add that in a follow-up commit once the core auth flow is confirmed working.

— Sam 🌊

jasonacox-sam:
Correction on my previous comment — the email-based cache selection fix was already applied in commit `47c1035` (see my May 13 comment above). The code now correctly does `if email and email in cache: cached_email = email` with fallback to the first entry when no email is specified. That follow-up comment was from a session that missed the earlier work. Everything is in place on main.

— Sam ⚡

jasonacox:
View at:
https://pypi.org/project/pypowerwall/0.15.7/

PR #313, opened 2026-06-09
Fix WiFi backoff race condition in multi-threaded proxy

### LINKED ISSUE #310: v1r mode not detecting "follower" PW3
Hi,

Thanks for all your work on this project, and especially with the work involved getting v1r mode working. I took the plunge and connected my powerwall 3 to my wired network a couple of weeks ago, and after a couple of false starts with the authentication, it has been working very reliably - much better than I could ever manage with the direct wifi connection.

I have been using pypowerwall-proxy configured in v1r + wifi backup mode. One issue that I have noticed is that when the wifi connection is active my dual PW3 setup seems to be correctly detected and I can see the details for both units under /pods and /vitals.

When the wifi connection drops out and I am querying by v1r alone, I can no longer see the second "follower" PW3 at either of these endpoints. I've also confirmed this behaviour by running in v1r mode only.

Thanks again.

jasonacox-sam:
## Problem

Fixes #310 — the WiFi exponential backoff for follower TEDAPI queries jumps to the ~2h maximum in a single burst instead of progressing gradually (60s → 120s → 240s → ...).

**Root cause:** Multiple concurrent threads (one per endpoint poller) all call `_post_tedapi_wifi()` at the same time during a WiFi outage. They all pass the cooldown check simultaneously (none has set it yet), then all increment `wifi_fail_count` independently in the same second:

```
fail_count goes: 1 → 2 → 3 → 4 → 5 → 6 → 7  (7 threads)
backoff goes:   120s → 240s → 480s → 960s → 1920s → 3840s → 7680s (capped)
```

This produces the ~6426s cooldown the reporter saw in `/health`.

## Fix

Add a `threading.Lock` (`_wifi_lock`) that protects all reads and writes of `wifi_fail_count` and `wifi_cooldown` in `_post_tedapi_wifi()`. The cooldown check is now atomic with the cooldown set — only one thread at a time can increment the failure counter and compute the backoff.

**Before:** 7 concurrent failures → backoff jumps straight to 7680s (~2h)
**After:** 7 concurrent failures → first sets 120s, remaining 6 see cooldown active and return immediately

## Testing

- Module imports cleanly
- No new dependencies (threading already imported)
- Lock is per-instance (no cross-instance contention)

### DIFF
--- a/pypowerwall/tedapi/__init__.py
+++ b/pypowerwall/tedapi/__init__.py
@@ -1384,11 +1384,13 @@ class TEDAPI:
             self._test_wifi_path()
             if not self.wifi_available:
                 with self._wifi_lock:
-                    self.wifi_fail_count += 1
-                    backoff = min(60 * (2 ** self.wifi_fail_count), 7680)  # caps at ~128 min
-                    self.wifi_cooldown = time.time() + backoff
-                    log.debug("WiFi unavailable, next retry in %.0fs (failure #%d)",
-                               backoff, self.wifi_fail_count)
+                    # Re-check cooldown: another thread may have set one while we tested
+                    if self.wifi_cooldown <= time.time():
+                        self.wifi_fail_count += 1
+                        backoff = min(60 * (2 ** self.wifi_fail_count), 7680)  # caps at ~128 min
+                        self.wifi_cooldown = time.time() + backoff
+                        log.debug("WiFi unavailable, next retry in %.0fs (failure #%d)",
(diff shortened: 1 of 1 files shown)


jasonacox-sam:
Good catch, Copilot. All four failure paths now re-check `wifi_cooldown` under the lock before incrementing `wifi_fail_count` — if another thread already activated a cooldown during the request window, we leave the existing backoff in place instead of stacking another increment on top.

Pushed in dcdc630. CI should pick it up shortly.

— Sam 🌊

PR #351, opened 2026-07-05
fix(windows): cap TLS to 1.2 on Windows for Tesla auth (#350)

jasonacox-sam:
## Summary

Fixes the Windows `403 Forbidden` issue reported in #350 where Tesla rejects tokens obtained via the PKCE code exchange on native Windows Python.

### Root Cause

Windows Python bundles an OpenSSL build whose **TLS 1.3 ClientHello fingerprint** is rejected by Tesla during:
- PKCE authorization code exchange (`_exchange_code()`)
- Token refresh (`_refresh_access_token()`)

The tokens come back "tainted" — they work for `auth.tesla.com` endpoints but return `403 Forbidden` on every `owner-api.teslamotors.com/api/1/*` call. This is the same class of TLS fingerprint issue we fixed for Alpine/musl libc Docker images in #345, but specific to native Windows.

### Changes

**1. Platform-aware TLS (`tesla_auth.py`, `teslapy/__init__.py`)**
- `_httpx_auth_verify()` now sets TLS 1.2 as the floor (HTTP/2 minimum) on all platforms
- On Windows (`sys.platform == 'win32'`), caps `maximum_version` to TLS 1.2 to avoid the rejected fingerprint
- On Linux/macOS, retains TLS 1.3 pinning (existing behaviour, works correctly)

**2. HTTP/2 for register command (`v1r_register.py`)**
- `api_call()` now uses `httpx` with HTTP/2 for Tesla API endpoints (was `urllib`/HTTP/1.1)
- Tesla now requires HTTP/2 for `owner-api.teslamotors.com/api/1/*` calls (June 2026)
- Falls back to urllib for non-Tesla URLs or if httpx is not installed

**3. Documentation (`README.md`)**
- Added Windows troubleshooting callout with three alternatives:
  - tesla_auth desktop app (generates tokens without Python TLS issues)
  - Full Linux VM (Hyper-V Ubuntu) — not WSL2 (WebKitGTK does not work in WSL2)
  - Remote authtoken from another machine

**4. Diagnostics (`__main__.py`)**
- `cloudcheck` now shows platform-aware TLS info on Windows

### Testing

- All modified files pass `py_compile`
- TLS context verified: Linux gets TLS 1.2–1.3 range, Windows would get TLS 1.2 cap
- No existing behaviour changed on Linux/macOS
- Test suite errors are pre-existing (protobuf version mismatch, unrelated to this PR)

### Context

- Issue #350 — @AndrewTapp's excellent bug report with full `cloudcheck -debug` output
- PR #345 — the Alpine/musl libc Docker fix (same class of TLS fingerprint issue)
- WSL2 cannot run the WebView auth because WebKitGTK requires a full desktop stack

### Does this need merge approval?

Yes — this is a `jasonacox` repo. @jasonacox please review.

— Sam ⚡

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,13 @@ python3 -m pypowerwall authtoken
 python3 -m pypowerwall setup -headless
 ```
 
+> ⚠️ **Windows Users:** Native Windows Python bundles an OpenSSL build whose TLS 1.3 ClientHello fingerprint is rejected by Tesla during the PKCE token exchange. The token comes back "tainted" — `setup` and `register` fail with `403 Forbidden` on `owner-api.teslamotors.com`. **As of v0.16.1+, pypowerwall automatically caps TLS to 1.2 on Windows** to avoid this. If you still see 403 errors, use one of these alternatives:
+> * **[tesla_auth](https://github.com/adriankumpf/tesla_auth/releases) desktop app** — generates RT+AT tokens on Windows without Python TLS issues. Copy both tokens and use `setup -headless`.
+> * **Full Linux VM** (e.g. Hyper-V Ubuntu) — not WSL2, which lacks the GTK/WebKit stack needed for the WebView login window.
+> * **Another machine** (Mac, Linux, RPi) — run `authtoken` there and paste the tokens into `setup -headless` on Windows.
+>
+> See [issue #350](https://github.com/jasonacox/pypowerwall/issues/350) for full background.
+
 > 💡 **Alternative:** The [tesla_auth](https://github.com/adriankumpf/tesla_auth/releases) desktop app (Mac/Windows/Linux) can also generate tokens. For iOS users, the [Auth app for Tesla](https://apps.apple.com/us/app/auth-app-for-tesla/id1552058613) provides tokens directly on-device.
 
 ### TEDAPI Mode - Option 4
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,18 @@
 * note(tedapi): the default `V2024_06` path and its protobufs are unchanged — the library floor stays at `protobuf>=4.25.1` (the `local`/`V2024_06` `*_pb2.py` are still generated with the protobuf 4.25.x toolchain and are byte-identical to prior releases). Only the opt-in `V2026_06` query set is built with the latest protoc, so its `*_pb2.py` embed a `runtime_version.ValidateProtobufRuntimeVersion()` guard and require `protobuf>=6.33.6`. Those modules are imported lazily only when `tedapi_api_version="V2026_06"` is selected, so an older runtime raises a clear, actionable error (`tedapi_api_version="V2026_06" requires protobuf>=6.33.6 — pip install -U protobuf`) at opt-in time — everyone on the default path is unaffected.
 * robustness(tedapi): when the gateway rejects a `V2026_06` signed query, pyPowerwall now logs a warning suggesting a fallback to `tedapi_api_version="V2024_06"` — a total `V2026_06` failure after a firmware update most likely means Tesla rotated the query signing keys, so the bundled signatures no longer validate.
 
+## v0.16.1 - Windows TLS Fix for Tesla Auth
+
+* fix(windows): cap TLS to 1.2 on Windows to avoid Tesla token fingerprint rejection (#350)
+  * Windows Python bundles an OpenSSL build whose TLS 1.3 ClientHello fingerprint is rejected by Tesla during PKCE code exchange and token refresh — tokens come back "tainted" and all `owner-api.teslamotors.com` calls return `403 Forbidden`
+  * `_httpx_auth_verify()` in `tesla_auth.py` and `teslapy/__init__.py` now caps `maximum_version` to TLS 1.2 on Windows (`sys.platform == 'win32'`); Linux/macOS retain TLS 1.3 pinning
+  * TLS 1.2 fully supports HTTP/2, so ALPN negotiation and h2 framing are unaffected
+* fix(register): `api_call()` in `v1r_register.py` now uses httpx with HTTP/2 for Tesla API endpoints (was urllib/HTTP/1.1)
+  * Tesla now requires HTTP/2 for `owner-api.teslamotors.com/api/1/*` calls (June 2026)
+  * Falls back to urllib for non-Tesla URLs or if httpx is not installed
+* docs(windows): README updated with Windows setup troubleshooting — native Windows TLS, WSL2 WebKitGTK limitation, and alternatives (tesla_auth app, full Linux VM, remote authtoken)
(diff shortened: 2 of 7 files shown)


jasonacox:
@jasonacox-sam  a few problems here:

* We don't know that TLS v1.2 will work on Windows - needs testing
* We went through quite a bit to get to v1.3 for MacOS and Linux - did we confirm that v1.2 does not work on those platforms? If so, we need to change the logic here to enforce v1.3 for MacOS and Linux.

Take a look at the reviews above as well.

jasonacox-sam:
@jasonacox — both concerns are valid. Here's where things stand:

**1. Copilot is right — the non-Windows minimum_version was a regression**

The previous commit set `minimum_version = TLSv1_2` for all platforms, which allowed TLS 1.2 negotiation on macOS/Linux even though the intent was TLS 1.3 only. Copilot caught this across all three affected files (`tesla_auth.py`, `teslapy/__init__.py`, `v1r_register.py`).

Just pushed a fix: non-Windows now gets `min = max = TLSv1_3` (strict pin, same as before this PR). Windows stays `min = max = TLSv1_2`. The platform split is now clean and symmetric.

**2. Windows TLS 1.2 has not been validated yet**

You're right to flag this. The root cause analysis in #350 strongly suggests the Windows OpenSSL TLS 1.3 fingerprint is the culprit — but we haven't had confirmation that capping to TLS 1.2 actually fixes it on a real Windows machine. @AndrewTapp said at the end of the #350 thread that he's happy to test native Windows when he returns at the end of the week.

My recommendation: hold off merging until Andrew tests. If TLS 1.2 doesn't solve it on Windows, we need to know before this ships. The PR is in good shape on the non-Windows side now; just needs that one data point.

— Sam ⚡

jasonacox:
@jasonacox-sam Please review and resolve the merge conflicts so we can merge.

jasonacox-sam:
Rebased and cleaned up — ready for merge. Here's what I did:

**Merge conflict (RELEASE.md)**
The conflict was between the upstream `Unreleased - TEDAPI V2026_06` section added by PR #340 and v0.16.1 notes from this branch. Resolved by keeping both sections in chronological order: TEDAPI Unreleased first (newest upstream work), then `v0.16.1` below it.

**Copilot review — TLS version settings (all three files)**
The three inline comments on `tesla_auth.py`, `teslapy/__init__.py`, and `v1r_register.py` were already addressed in the `fix(tls)` commit from July 5 — non-Windows now pins `min=max=TLSv1_3` as intended, and Windows keeps `min=max=TLSv1_2`. Verified those are correct post-rebase.

**Copilot review — misleading `SSL_CTX` comment + usage (v1r_register.py)**
The 4th comment pointed out that `SSL_CTX` (an unpinned `ssl.create_default_context()`) was being described as for "non-Tesla urllib calls" but was actually used in `step2_exchange_token()` against `fleet-auth.prd.vn.cloud.tesla.com` (a Tesla endpoint) and as the urllib fallback in `api_call()`. Fixed both:
- Replaced `SSL_CTX` with `_ssl_ctx()` in `step2_exchange_token()` and the `api_call()` urllib fallback, so the platform-aware TLS policy (TLS 1.2 on Windows, TLS 1.3 elsewhere) applies consistently to every outbound call
- Removed the now-unused `SSL_CTX` constant entirely

**Tests:** 112 unit tests pass, 53 proxy tests pass.

Thanks to @AndrewTapp for the detailed debugging and Windows testing that made this fixable. 🙏

— Sam ⚡

jasonacox:
@jasonacox-sam One thing I'm wondering - for the non-Windows OS use case:

v1r_register.py's new _ssl_ctx() helper is platform-branched for which TLS version to use (1.2 on Windows, 1.3 elsewhere), but that's a new restriction on non-Windows. Previously this file used a bare ssl.create_default_context() with no version pin at all — now it forces strict TLS 1.3-only on Linux/macOS. If any of those environments can't negotiate TLS 1.3 (older OpenSSL), register/setup would now fail there when it didn't before. 

Just to be clear, is the intent to force TLS 1.3 or fail on MacOS or Linux? If so, this looks good. If not, that may be some logic we want to adjust.

jasonacox-sam:
@jasonacox — yes, the intent is to force TLS 1.3 on macOS/Linux, and I believe this is safe for a few reasons:

1. **It's already the existing behavior in `tesla_auth.py`** — the `_httpx_auth_verify()` function there has been pinning to TLS 1.3 on non-Windows since before this PR. The new `_ssl_ctx()` in `v1r_register.py` deliberately mirrors that same policy to make the two files consistent. So this isn't a new restriction for anyone who was already going through the httpx auth path.

2. **The `elif hasattr(ssl.TLSVersion, 'TLSv1_3')` guard is a graceful fallback** — if the runtime doesn't have TLS 1.3 support (very old OpenSSL), neither branch fires and `ctx` returns as a bare `ssl.create_default_context()`. No regression for ancient environments.

3. **TLS 1.3 is effectively universal on pypowerwall's supported runtime** — Python 3.7+ ships with OpenSSL 1.1.1+ (which includes TLS 1.3), and Tesla's auth endpoints are known to accept TLS 1.3 on Linux/macOS (that's been the live code path in tesla_auth.py all along).

If you wanted a more permissive non-Windows behavior — setting only `ctx.minimum_version = ssl.TLSVersion.TLSv1_2` and no maximum, so TLS 1.2 or 1.3 both negotiate — that would be a reasonable conservative adjustment. But given `tesla_auth.py` has been running strict TLS 1.3 on non-Windows without reports of breakage, I'd lean toward keeping the current behavior for consistency. Your call on whether you want that extra headroom or prefer the cleaner pin.

— Sam ⚡

jasonacox:
@jasonacox-sam Sold! Thanks for holding your ground on this and the well explained basis.  Please create a new tag and github release for this version. Also, please drop a post on X.

View at:
https://pypi.org/project/pypowerwall/0.16.1/

Proxy:
jasonacox/pypowerwall:0.16.1t95

jasonacox:
@jasonacox-sam  see thread.

jasonacox-sam:
@jasonacox — done! Here's what's been completed:

**✅ Git tag created:** `v0.16.1` → commit `f2897a4` (the `fix(requirements)` bump on main)

**✅ GitHub release published:** [v0.16.1 — Windows TLS Fix for Tesla Auth](https://github.com/jasonacox/pypowerwall/releases/tag/v0.16.1)

**✅ X post published:** [pypowerwall v0.16.1 on @jasonacox_sam](https://x.com/jasonacox_sam/status/2076460826949795852)

> _Note: the earlier `403 Forbidden` from `x_post.py` was not a credentials or permissions issue — the tweet text was over X's 280-character limit, which returns a generic 403. Shortened it and posted successfully._

— Sam ⚡

PR #358, opened 2026-07-12
fix(proxy): add built-in HEALTHCHECK using curl instead of wget

### LINKED ISSUE #346: v5.1.2 unhealthy - wget missing
Reporting a new issue for container health reporting as "unhealthy" after 5.1.2 upgrade. Error below.

```/bin/sh: 1: wget: not found```

### PR DESCRIPTION
**Pull request by jasonacox-sam**
## Summary

Fixes #346

The proxy Docker image switched its base from Alpine to Debian-slim in t94 (PR #345) to fix Tesla TLS token refresh 403 errors. Alpine includes `wget` by default; Debian-slim does not. Many compose files (including older Powerwall-Dashboard releases) reference `wget` in their healthcheck definition, causing:

```
/bin/sh: 1: wget: not found
```

and the container reporting `unhealthy`.

## Fix

Add a `HEALTHCHECK` instruction directly to both `Dockerfile` and `Dockerfile.beta` using `curl` (already installed in the image) hitting the `/health` endpoint:

```dockerfile
HEALTHCHECK --interval=30s --timeout=10s --start-period=60s --retries=3 \
    CMD curl -sf http://localhost:8675/health > /dev/null || exit 1
```

This gives the image a working default healthcheck **built into the image itself**, so it works regardless of what compose file references it. Compose-level healthchecks still override this if defined.

Also updated the stale comment that still referenced `wget` (now says `curl`).

This mirrors the fix already applied in Powerwall-Dashboard's `powerwall.yml`, but moves the fix into the image itself for broader coverage.

## Testing

- [x] HEALTHCHECK syntax validated
- [x] `/health` endpoint confirmed present in server.py
- [x] `curl` confirmed in both Dockerfiles's install step
- [ ] Docker image build + healthcheck verification (will test after merge)

— Sam ⚡

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -19,7 +19,7 @@ WORKDIR /app
 
 # Install system dependencies (package manager differs by base image).
 # tini is a tiny init that, as PID 1, reaps orphaned child processes (e.g. the
-# wget spawned by the container HEALTHCHECK) and forwards signals. This is
+# curl spawned by the container HEALTHCHECK) and forwards signals. This is
 # defense-in-depth alongside the SIGCHLD reaper in server.py, and also covers
 # any future child processes regardless of how the operator runs the image.
 RUN if command -v apk > /dev/null; then \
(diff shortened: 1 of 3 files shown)


jasonacox-sam:
Closing in favour of #357, which has the same Dockerfile fix plus the `server.py` build bump. I've pushed the `RELEASE.md` t96 entry from this PR onto #357's branch so nothing is lost. #357 is now the complete fix for #346.

PR #364, opened 2026-07-26
fix(tedapi): Correct Content-type typo, octet-string -> octet-stream

Nexarian:
# Summary

`application/octet-string` is not a media type. The binary-payload type is `application/octet-stream` (RFC 2046); the TEDAPI request headers have carried the typo since the endpoint was first added.

Replaces all 23 occurrences across the library, the tools/tedapi scripts, and their documented curl examples.

This is a correctness and consistency fix rather than a behavior change: the gateway evidently does not validate the header — the bogus value has been accepted for years without issue. Fixing the typo means the documented curl commands can be pasted somewhere stricter without mystifying failures, and the repo is no longer inconsistent with itself.

No functional change intended; no test asserted on the old value.

**Hardware validation (PW3, fw 26.18.3):** All TEDAPI calls (`get_config`, `get_status`, `vitals`, `get_firmware_version`) return identical payloads with the corrected header across both the primary WiFi path and the v1r WiFi fallback path. See [test results](https://github.com/jasonacox/pypowerwall/pull/364#issuecomment-5153220265).

Note: This is also a "parted out" change from https://github.com/jasonacox/pypowerwall/pull/359

# Supposition

How was this introduced? The likely confusion is real and understandable: OCTET STRING is a genuine ASN.1 universal type (tag 4, used throughout X.509/DER). Someone deep in binary-protocol work reaches for "octet string" from that vocabulary while writing an HTTP header. Fittingly, the Tesla bundle contains ASN.1 code with exactly that terminology — "Multi-octet tag encoding unsupported", tagClassByName["universal"] — so both words live in this problem space.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,5 +1,9 @@
 # RELEASE NOTES
 
+## Unreleased
+
+* fix(tedapi): correct Content-type request header typo `application/octet-string` → `application/octet-stream` (RFC 2046); gateway does not validate the header (#364)
+
 ## v0.16.2 - TEDAPI Fallback, v1r Diagnostics, and Firmware Version Improvements
 
 * feat(proxy): TEDAPI SolarOnly fallback mode — when TEDAPI connectivity is lost, the proxy automatically continues serving solar data without interruption. Enabled via `PW_TEDAPI_RECOVERY=yes`. (#361)
--- a/pypowerwall/tedapi/__init__.py
+++ b/pypowerwall/tedapi/__init__.py
@@ -1269,7 +1269,7 @@ class TEDAPI:
             session.headers.update({'Connection': 'close'})  # This disables keep-alive
         session.verify = False
         session.auth = ('Tesla_Energy_Device', self.gw_pwd)
-        session.headers.update({'Content-type': 'application/octet-string'})
+        session.headers.update({'Content-type': 'application/octet-stream'})
         return session
 
     def _init_wifi_session(self, gw_pwd: str):
@@ -1288,7 +1288,7 @@ class TEDAPI:
             session.headers.update({'Connection': 'close'})
         session.verify = False
         session.auth = ('Tesla_Energy_Device', gw_pwd)
-        session.headers.update({'Content-type': 'application/octet-string'})
+        session.headers.update({'Content-type': 'application/octet-stream'})
         self.wifi_session = session
         log.debug(f"WiFi fallback session initialized for {self.wifi_host}")
 
--- a/tools/tedapi/ComponentsQuery.py
+++ b/tools/tedapi/ComponentsQuery.py
@@ -58,7 +58,7 @@ pb.message.config.send.file = "config.json"
 pb.tail.value = 1
 url = f'https://{GW_IP}/tedapi/v1'
 r = requests.post(url, auth=('Tesla_Energy_Device', gw_pwd), verify=False,
-    headers={'Content-type': 'application/octet-string'},
+    headers={'Content-type': 'application/octet-stream'},
     data=pb.SerializeToString(), timeout=5)
 #print(f"Response Code: {r.status_code}")
 # Decode response
@@ -90,7 +90,7 @@ pb.message.firmware.request = ""
 pb.tail.value = 1
 url = f'https://{GW_IP}/tedapi/v1'
 r = requests.post(url, auth=('Tesla_Energy_Device', gw_pwd), verify=False,
-    headers={'Content-type': 'application/octet-string'},
+    headers={'Content-type': 'application/octet-stream'},
     data=pb.SerializeToString(), timeout=5)
 print(f"Response Code: {r.status_code}")
 # write raw response to file
@@ -118,7 +118,7 @@ pb.message.payload.send.b.value = "{\"pwsComponentsFilter\":{\"types\":[\"PW3SAF
 pb.tail.value = 1
 url = f'https://{GW_IP}/tedapi/v1'
 r = requests.post(url, auth=('Tesla_Energy_Device', gw_pwd), verify=False,
-    headers={'Content-type': 'application/octet-string'},
+    headers={'Content-type': 'application/octet-stream'},
     data=pb.SerializeToString(), timeout=5)
 print(f"Response Code: {r.status_code}")
 # Decode response
@@ -155,7 +155,7 @@ for battery in battery_blocks:
     pb.tail.value = 1
     url = f'https://{GW_IP}/tedapi/v1'
     r = requests.post(url, auth=('Tesla_Energy_Device', gw_pwd), verify=False,
-        headers={'Content-type': 'application/octet-string'},
+        headers={'Content-type': 'application/octet-stream'},
         data=pb.SerializeToString(), timeout=5)
     print(f"Response Code: {r.status_code}")
     if r.status_code == 200:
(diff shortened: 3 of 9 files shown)


jasonacox-sam (approved):
The correction is right — `application/octet-stream` is the IANA-registered binary type (RFC 2046 §4.5.1); `application/octet-string` is not a media type. The Tesla gateway has been lenient enough to accept the typo, so this is a cleanup rather than a bug fix for end-users, but it's the kind of thing that quietly misfires with stricter proxies or future firmware.

All 23 occurrences updated consistently across the library, the tools scripts, and the README example — no misses I can find. Clean and ready.

— Sam ⚡

jasonacox:
Thanks @Nexarian — good catch, and good discipline parting this out of #359. The fix is correct in intent (`application/octet-string` is indeed not a media type), the replacement is complete for every tracked file (the only remaining occurrences are in untracked local scratch), and the origin story checks out (`c319ec6`, Nov 2023 — mea culpa 😄).

Before this merges, though, one factual correction and two asks.

## The safety claim needs correcting

The PR description says octet-stream is proven safe because "tedapi_v1r.py sends `application/octet-stream` to the same endpoint today." That's not quite right — I checked every in-repo `octet-stream` user:

- `tedapi_v1r.py` → posts to **`/tedapi/v1r`**
- `tools/tedapi-lan/lan_tedapi.py` → also **`/tedapi/v1r`**
- `pwsimulator` → our own stub; accepts anything, proves nothing about Tesla firmware

The two changed library lines (`_init_session` / `_init_wifi_session`) talk to **`/tedapi/v1`** — a different endpoint, plausibly a different handler on the gateway. So there is currently no in-repo hardware evidence that `/tedapi/v1` accepts `octet-stream`. The closest real-world data point is your own #359 bearer testing, which did hit `/tedapi/v1` with `octet-stream` — on your solar-only gateway firmware. PW2/PW3 remain unverified.

Is it *likely* fine? Yes — a server that has ignored a bogus value for years is almost certainly ignoring the header entirely. But this change has zero functional upside on hardware that ignores the header, and its failure mode is the primary transport for every TEDAPI/hybrid user and every Powerwall-Dashboard PW3 install running 24/7. That asymmetry means we verify before we merge, not after.

**Please update the PR description** so the justification reads "the gateway evidently doesn't validate the header (bogus value accepted for years)" rather than the v1r equivalence — future archaeology will thank you.

## @jasonacox-sam — hardware validation needed

Please run this branch against the PW3 (fw 26.18.x) and report results here:

1. **WiFi TEDAPI** (192.168.91.1, `gw_pwd` basic auth): `get_config()`, `get_status()`, `vitals()`, `get_firmware_version()` — exercises the `_init_session` change
2. **v1r with `wifi_host` set**, forcing the WiFi fallback path (drop the LAN route so `lan_failed` engages): one config + one status fetch — exercises the `_init_wifi_session` change
3. Confirm HTTP response codes and payload parity vs. the same calls on `main`

If anyone watching has a **PW2/PW+** and can run item 1 against this branch, that would close the last gap — PW2 firmware is the oldest and most likely to behave differently. (My PW2 can cover this if no volunteer turns up first.)

## RELEASE.md

Two lines of shipped library code changed, so this needs an entry under Unreleased — one bullet is plenty:

```
* fix(tedapi): correct Content-type request header typo `application/octet-string` → `application/octet-stream` (RFC 2046); gateway does not validate the header (#364)
```

With the description fixed, the RELEASE entry added, and Sam's matrix green, this merges. Nice find.

jasonacox-sam:
Jason — ran the full matrix against the local PW3 (fw **26.18.3**, build `184289b9`). All green. ✅

## Test 1 — WiFi TEDAPI direct (`192.168.91.1`, basic auth `_init_session`)

| Call | Result | Payload |
|------|--------|---------|
| `get_din()` | ✅ Pass | `<DIN>` returned |
| `get_firmware_version()` | ✅ Pass | `26.18.3 184289b9` |
| `get_config()` | ✅ Pass | 2816 bytes |
| `get_status()` | ✅ Pass | ~18.5 KB, top-level keys: `control`, `esCan`, `neurio`, `pw3Can`, `system` |
| `vitals()` | ✅ Pass | ~6.1 KB, full device hierarchy (`STSTSM`, `TESLA`, `TESYNC`, `TEPOD`, `PVAC`, `PVS`, `TEPINV`) |

## Test 2 — v1r WiFi fallback (`_init_wifi_session`, `lan_failed` forced)

Connected via hardwire (`_connect_v1r` succeeded, RSA key verified), then forced `lan_failed = True` to route primary queries through the WiFi fallback session:

| Call | Result | Payload |
|------|--------|---------|
| `get_config()` | ✅ Pass | 2816 bytes |
| `get_status()` | ✅ Pass | ~18.5 KB, same top-level structure |

The WiFi session's `Content-type` header was confirmed as `application/octet-stream` (the fix).

## Payload parity vs. `main` (`octet-string`)

| Endpoint | `main` size | PR size | Parity |
|----------|------------|---------|--------|
| WiFi `get_config()` | 2816 B | 2816 B | **Exact hash match** ✅ |
| WiFi `get_status()` | ~18.5 KB | ~18.5 KB | ✅ (sizes within 0.1%, expected drift — live data) |
| WiFi `vitals()` | ~6.1 KB | ~6.1 KB | ✅ (within 0.6%, live data) |
| WiFi `get_firmware_version()` | `26.18.3 184289b9` | `26.18.3 184289b9` | ✅ Identical |
| Fallback `get_config()` | 2816 B | 2816 B | **Exact hash match** ✅ |
| Fallback `get_status()` | ~18.5 KB | ~18.5 KB | ✅ (within 0.1%, live data) |

`get_config()` hashes are identical between branches (config is static), confirming the header change has zero functional impact on the gateway's response. Live endpoints (`get_status`, `vitals`) show expected minor size drift from sampling at different moments.

**Bottom line:** the PW3 gateway on firmware 26.18.3 accepts `application/octet-stream` identically to `application/octet-string` across both the primary WiFi path and the v1r WiFi fallback path. No HTTP errors, no payload differences, no regressions.

---

### Reminders before merge

@Nexarian — two asks from Jason's review:
1. **Update the PR description** to drop the v1r-equivalence justification and instead note "the gateway evidently doesn't validate the header (bogus value accepted for years)"
2. **Add a RELEASE.md entry** under Unreleased:
   ```
   * fix(tedapi): correct Content-type request header typo `application/octet-string` → `application/octet-stream` (RFC 2046); gateway does not validate the header (#364)
   ```

Clean find, clean fix. Thanks @Nexarian — and good discipline splitting this out of #359.

— Sam ⚡

jasonacox:
@jasonacox-sam  Can you handled the two remaining items? Once that is done, you are approved to merge.

jasonacox-sam:
Jason — both items handled.

**1. PR description updated:** Removed the v1r-equivalence justification. The Summary now reads:

> This is a correctness and consistency fix rather than a behavior change: the gateway evidently does not validate the header — the bogus value has been accepted for years without issue.

Also added a hardware-validation note linking to the [PW3 test results](https://github.com/jasonacox/pypowerwall/pull/364#issuecomment-5153220265) above.

**2. RELEASE.md entry added** under a new `## Unreleased` section:

```
* fix(tedapi): correct Content-type request header typo `application/octet-string` → `application/octet-stream` (RFC 2046); gateway does not validate the header (#364)
```

Pushed to the PR branch (commit `2083653`), squashed and merged. ✅

— Sam ⚡

jasonacox:
Thanks @Nexarian ! Good update.

PR #379, opened 2026-09-07
fix(tedapi): enable PW3 v1r islanding commands

mcmahonl19:
## Summary

- Make the existing `Powerwall.go_off_grid(confirm=True)` and
  `Powerwall.reconnect_grid()` methods work through the PW3 v1r/TEDAPI backend.
- Send Tesla's signed legacy `setIslandMode` command directly; no public API,
  proxy, FleetAPI, or configuration-write behavior changes.

## Validation

- Live-hardware validated on a Powerwall 3 using v1r/TEDAPI.
- Added non-actuating unit tests for the off-grid/reconnect wire formats and
  TEDAPI routing.


jasonacox-sam:
Thanks for the PR — nice to see islanding control come to v1r! The scope looks tight: routing the existing `go_off_grid()`/`reconnect_grid()` calls through the signed `setIslandMode` path with no changes to proxy/FleetAPI behavior, plus non-actuating unit tests.

I have a live PW3 setup (dual PW3 + expansion, fw 26.18.1) with a registered v1r key, so I'll run a proper review plus hardware validation on the wire format and routing shortly. A couple of things I'll check in particular:

- That `set_island_mode` failures surface properly (not swallowed) — control commands that silently no-op are worse than loud errors
- Interaction with the existing config-write path in `write_config_v1r()` (no lock or session contention)
- That the command is gated behind `confirm=True` semantics the same way the cloud path is

— Sam ⚡

mcmahonl19:
Thanks, Sam — I appreciate the detailed review and live-hardware
validation.

The public confirm=True guard is intentionally unchanged; the v1r routing
is only reached after that existing gate. The current implementation
follows the TEDAPI backend’s None-on-failure pattern, but an explicit
control-command failure should not be mistaken for success. I’m happy to
adjust that behavior to match the project’s preferred control-command error
contract if your review finds it needs tightening.

I also kept this separate from write_config_v1r() deliberately, using the
signed command path only. I’ll watch for your hardware results and address
any findings promptly.

jasonacox-sam (approved):
Review complete — code read, unit tests run, and live hardware validation done.

## Live hardware validation ✅

Tested against live PW3 (dual Powerwall 3 + expansion, fw 26.26.4) with a registered v1r key, using the PR branch code (`sys.path` front-loaded, not the installed package):

| Check | Result |
|---|---|
| `go_off_grid(confirm=False)` guard | ✅ refused (`None`), no command sent |
| `go_off_grid(confirm=True)` → `setIslandMode(mode=6, force=True)` | ✅ `{'mode': 6, 'force': True, 'result': 1}` — contactor **opened**: GridStatus 1→0, grid flow 4092 W→0 W, home carried by solar + battery |
| `reconnect_grid()` → `setIslandMode(mode=1)` | ✅ `{'mode': 1, 'force': False, 'result': 1}` — contactor **closed**: GridStatus 0→1, grid flow restored |
| Unit tests (`test_v1r_islanding.py`, `test_powerwall_core.py`) | ✅ 41 passed |

The manual protobuf wire encoding is correct — legacy `setIslandModeRequest` at TEGMessages field 3, appended after `SerializeToString()` since the V2024_06 generated set omits it, and the response parser correctly extracts `setIslandModeResponse.result`. The physical behavior confirms the gateway accepts and acts on it.

## On the points from your last comment

- **`confirm=True` guard** — agreed and verified: the top-level gate refuses before the v1r routing is ever reached. No change needed.
- **Error contract** — the `None`-on-failure pattern matches the rest of the TEDAPI backend, so I'd keep it for consistency rather than introduce a new exception contract in this PR. One small suggestion (non-blocking, your call): a `result` of `1` currently means success on both commands, but the semantics of other values aren't documented anywhere. A one-line docstring note or a log line when `result` is present but non-1 (or absent, which you already log as a warning) would make future debugging much easier for anyone hitting a non-actuating gateway.
- **Separation from `write_config_v1r()`** — agreed; the signed command path is the right transport for this and keeps the config-write surface untouched.

Nice work — this closes a real gap (islanding previously required cloud/Fleet even when v1r was paired). Approving from my side; merge is @jasonacox's call as always.

— Sam ⚡

mcmahonl19:
Thanks, Sam.

Since your hardware validation passed and the logging suggestion is
non-blocking, my preference is to leave the implementation as-is rather
than create another review cycle for a minor diagnostic change. If Jason
wants that tightened before merge, I’m happy to address it.

Regards,

Lynn

jasonacox:
## Maintainer update pushed: `d2e159c` — islanding moved onto the vendored proto

@mcmahonl19 — first: thank you. The command discovery, the field-number archaeology, and the hardware validation in this PR are the hard part, and all of it survives unchanged. I've pushed a maintainer-edit commit that relocates *how* the command is encoded, and I want to be transparent about why, since Sam and I had initially said the hand-rolled approach was acceptable.

### What changed in `d2e159c`

- `TEGAPISetIslandModeRequest`/`TEGAPISetIslandModeResponse` now live in the vendored `tedapi_combined.proto` (legacy TEG oneof fields 3/4, with a provenance + hardware-validation comment), pb2 regenerated via the canonical `tools/gen_proto.sh` toolchain.
- `send_island_mode()` now routes through the existing `send_teg_message()` helper; the ~85-line hand protobuf codec (varint encode/read, field walker, response scanner) is deleted. Net: the PR shrinks by ~47 lines while keeping every behavior path (guards, `ValueError` on bad mode, None-on-failure, missing-result warning).
- Added the diagnostic Sam suggested: a non-1 `result` now logs a warning ("1 is the observed success value").
- Nits: redundant `v1r` guard in `pypowerwall_tedapi.py` dropped (inner layer already guards); stale "cloud-only" islanding note in the facade docstring updated.
- Housekeeping: `v0.17.2` shipped from main while this PR was open, so this is retitled/bumped to **`v0.17.3`** in RELEASE.md and `version_tuple`.
- Discovered along the way: `gen_proto.sh` used GNU-only `sed -i -E`, which BSD/macOS sed misparses (corrupting the V2026_06 pb2 import rewrite on any Mac, including via the pre-commit hook). Fixed with portable `perl -pi -e`.
- Policy so the next contributor doesn't have to rediscover any of this: new AGENTS.md section ("A new TEDAPI protobuf message or TEG command") and a `README.md` in `pypowerwall/tedapi/protobuf/` covering which proto set a change belongs in (V2024_06 vs V2026_06), the dual-toolchain regeneration, and the byte-pinning-test convention.

### Why we changed course

Honest accounting: the review initially accepted the hand codec because we assumed extending the vendored proto was expensive (pb2 churn, toolchain pinning, re-validation risk). Jason pushed back with "why not do the right thing now?" — so we measured instead of assuming, and the assumed costs didn't hold: the pb2 diff is one blob line plus a mechanical offset table, `gen_proto.sh` already existed as the canonical regeneration path, and — decisively — **the generated encoder produces byte-for-byte identical wire output to your hand-rolled encoding**. Your own unit-test byte assertions pass *unchanged* against the new implementation; they now pin the generated encoder to the hardware-validated bytes, which is exactly the regression guard we want. Identical bytes also means your live PW3 validation carries over in full.

The principle now written into AGENTS.md: the vendored `.proto` files are the single source of truth for the TEDAPI wire schema — "the generated code omits the field" is a reason to extend the proto, not to bypass it. This was the third TEG command (max backup, cancel, islanding), so the pattern was worth settling now, on the PR that surfaced it.

### Verification on `d2e159c`

- Full non-live suite: **643 passed**; pylint unchanged (9.95/10); CLI reports 0.17.3.
- Wire bytes verified byte-identical on every path: off-grid (mode 6 + force), reconnect (mode 1, force omitted), missing-result → `result: None` + warning, transport failure → `None`, bad mode → `ValueError`.
- pb2 regenerated with the pinned 4.25.x legacy toolchain (matches the `check-protobuf.yml` CI check), not bare protoc.

### @jasonacox-sam — final hardware gate before merge

We are **not merging yet**. Even with byte-identical encoding proven at the unit level, this commit changed the encode path of a physical contactor command, so please re-run your PW3 validation matrix on `d2e159c` as the final gate:

1. `go_off_grid(confirm=False)` → refused, nothing sent
2. `go_off_grid(confirm=True)` → contactor opens (GridStatus 1→0, grid flow →0), returns `{'mode': 6, 'force': True, 'result': …}`
3. `reconnect_grid()` → contactor closes (GridStatus 0→1, grid flow restored)
4. New this round: confirm the `result` values you observe, and that any non-1/absent result produces the new warning log line rather than silence

Post the parity table when done; merge is my call after your green light.

jasonacox-sam:
## Hardware validation on `d2e159c` — full matrix ✅

@jasonacox — done, re-ran the validation matrix against the live system (dual PW3 + expansion, fw 26.18.1) on `d2e159c` (`0.17.3`). Green light from me.

| Step | Command | Result | Physical evidence |
|---|---|---|---|
| 1. Guard | `go_off_grid(confirm=False)` | `None` + log line `go_off_grid requires confirm=True` | No state change — solar/battery/level identical before & after |
| 2. Off-grid | `go_off_grid(confirm=True)` | `{'mode': 6, 'force': True, 'result': 1}` | GridStatus **1→0**, grid flow **→ 0.0 W**, battery picked up full home load (~1.9 kW discharge), solar unchanged |
| 3. Reconnect | `reconnect_grid()` | `{'mode': 1, 'force': False, 'result': 1}` | Contactor reclosed — GridStatus **0→1**, grid flow restored (−31 W → +75 W settling), battery discharge dropped back toward net |

**Item 4 — `result` values:** both commands returned `result: 1` on live hardware, matching the observed success value, so the new non-1 warning path wasn't exercised in this run — it's covered by the unit suite (missing-result → `None` + warning) but I want to be precise that no live non-1 result occurred to observe.

Behavior parity with the pre-refactor implementation is confirmed on-hardware: same return shapes, same guard semantics, and the encode-path change to the vendored proto produced accepted commands on both directions of the contactor. Ready for merge whenever you call it.

— Sam ⚡

jasonacox:
Pushed one more maintainer commit, `284b422` — **documentation for the new capability plus safety warnings**:

- **API.md** (Grid Island Control): documents the v1r local path (new in 0.17.3) alongside cloud/FleetAPI, the `{"mode", "force", "result"}` return shape and the `result=1` success convention, and adds a prominent ⚠️ warning: these commands physically operate the grid contactor, an islanded home with a depleted battery loses power, do not automate without handling the failure modes (script dies while off-grid, gateway unreachable for reconnect), and always verify grid status after a command instead of assuming success. The "v1r LAN Control" note also no longer implies islanding is a config-file write — it now distinguishes config writes (reserve/mode) from signed TEG commands (backup events, islanding).
- **README.md**: Functions list entries updated with the supported modes, plus the same extreme-care warning block below the list.
- **Facade docstrings + module Functions list** (`pypowerwall/__init__.py`): `go_off_grid`/`reconnect_grid` now carry the USE WITH EXTREME CARE warning at the top of their docstrings and appear in the module docstring's Functions list.

Docs-only — no code changes; suite still 643 passed, pylint 10.00 on the touched module.

@jasonacox-sam — the hardware-validation request from the previous comment still stands, now against `284b422` (the code under test is identical to `d2e159c`; only docs changed on top).

mcmahonl19:
Wow! Thanks for the quick turnaround. I really appreciate you taking the time to wrap this up.

jasonacox:
Thanks @mcmahonl19 ! This is a great addition to the library. Thank YOU! 🙏 

@jasonacox-sam Thanks for the hardware test validation! Please create the new tag for v0.17.3 and github release. Make sure to include all changes since last tag and contributors. Post on X.

View at:
https://pypi.org/project/pypowerwall/0.17.3/

Proxy container: jasonacox/pypowerwall:0.17.3t101

jasonacox-sam:
**v0.17.3 is released** 🚀 — tag pushed, GitHub release published, PyPI live, and the X announcement is up.

- **Release:** https://github.com/jasonacox/pypowerwall/releases/tag/v0.17.3
- **PyPI:** https://pypi.org/project/pypowerwall/0.17.3/
- **Proxy container:** `jasonacox/pypowerwall:0.17.3t101`
- **X:** https://x.com/jasonacox_sam/status/2097144932498890759

Lynn — thank you for this one. The command discovery, the TEGMessage field-number archaeology that got `setIslandMode` onto the wire, the signed-command implementation, and live PW3 validation in both contactor directions — the hard parts of this release were all yours, and they carried straight through to the final form: your byte-assertion unit tests now pin the vendored-proto encoder to the hardware-validated bytes, which is exactly the regression guard this capability deserves. Islanding control for v1r users wouldn't exist without this PR.

Full changelog and contributors are in the release notes above.

— Sam ⚡

PR #381, opened 2026-09-15
fix(tedapi): per-instance API locks so multi-gateway pollers don't serialize

Nexarian:
The @uses_api_lock getter locks were GLOBAL per function - one lock per method shared by every TEDAPI instance in the process. A server polling several gateways (one TEDAPI each) serialized all of them behind whichever gateway's fetch was in flight, and a slow gateway could starve the others into lock timeouts and stale-cache fallbacks.

The locks now live on the instance, keyed by method name (TEDAPI._api_lock()), so gateways never wait on each other while per-method serialization within one gateway is unchanged. acquire_lock_with_backoff() accepts the lock itself in addition to the legacy api_lock-holder convention; None stays a no-op.

This is a pre-req for enabling more robust multi-gateway support in pypowerwall-server


Nexarian:
@jasonacox-sam Please take a look!

pypowerwall/api_lock.py
```diff
@@ -44,14 +44,26 @@ def acquire_with_exponential_backoff(
...
+    lock as an ``api_lock`` attribute (legacy holder convention). ``None``,
+    or a holder without a lock, yields a no-op context.
     """
-    if lock_holder  is None or not hasattr(lock_holder , "api_lock"):
+    if lock_holder is None:
+        # no-op context
+        yield
+        return
+    if hasattr(lock_holder, "acquire") and hasattr(lock_holder, "release"):
+        lock: threading.Lock = lock_holder
```
jasonacox, line 60: Required (tiny): this annotation asserts something that isn't statically true — `lock_holder` is duck-typed here (it's whatever passed the `acquire`/`release` check), so annotating the assignment as `threading.Lock` is misleading to readers and type-checkers alike. Make it a plain assignment:

```python
lock = lock_holder
```

The docstring above already explains the duck-typing contract, which is the right place for it.

pypowerwall/tedapi/__init__.py
```diff
@@ -297,16 +309,34 @@ def _in_cooldown(self) -> bool:
...
+        (a getter called without the decorator, e.g. directly in tests) is
+        ``None`` out, which ``acquire_lock_with_backoff`` treats as "no lock".
+        """
+        if self_function is None:
+            return None
+        name = getattr(self_function, '__name__', None) or str(self_function)
+        with self._api_locks_guard:
+            lock = self._api_locks.get(name)
+            if lock is None:
+                lock = self._api_locks[name] = threading.Lock()
```
jasonacox, line 326: Required (tiny): use the same idiom we already use for exactly this pattern in `PyPowerwallCloud._site_api`, so the codebase has one way of doing per-name lazy locks:

```python
with self._api_locks_guard:
    return self._api_locks.setdefault(name, threading.Lock())
```

Same semantics, three lines shorter, and grep-consistent with the cloud backend.

jasonacox (changes requested):
Nice catch, @Nexarian — this is a real and subtle bug. A lock stored as a function attribute is one lock per method for the entire process, so a server polling several gateways really did serialize every gateway behind whichever fetch was in flight, with the starvation-into-stale-cache failure mode you describe. Good find, and thank you for bringing it with tests.

I did a full review pass on this and verified the safety claims independently:

**What I verified**

- **Only TEDAPI uses this machinery** — `acquire_lock_with_backoff` has no other in-repo callers, so no other backend silently loses locking when the decorator stops installing `func.api_lock`.
- **No in-repo path creates two `TEDAPI` instances for one gateway** (hybrid local and full-TEDAPI each construct exactly one), so the accidental cross-instance protection the global lock provided wasn't load-bearing. And I grepped pypowerwall-server: zero references to `api_lock`, so removing the function attribute breaks nothing downstream. **No breaking change** — same-gateway, same-method serialization is unchanged.
- **Thread safety of the fix itself**: lazy creation is correctly guarded (`_api_locks_guard`), and your 16-thread barrier test proves single-lock identity under race. The design also mirrors the locking pattern we already use in `PyPowerwallCloud._site_api` — right way to do it.
- **Lock identity is consistent**: the decorator injects the undecorated function while tests pass the `@wraps`-preserved wrapper; both resolve to the same `__name__`, same lock. Checked.
- **AGENTS.md concurrency rules**: real `threading.Lock`s ✅, and per-instance scoping moves us *closer* to the "never hold a lock across a network call" intent — the pre-existing hold-across-fetch design (bounded waiters, stale-serve) is unchanged and out of scope here.
- Tests: 16 new, all passing locally and in CI (fully green, coverage +0.40%). The suite covers the actual starvation scenario, per-method/per-instance identity, the race, the removed function attribute, and all four `acquire_lock_with_backoff` input shapes.

**Two small required items** (we fix everything in the PR that surfaces it — no deferred nits in this project): both are one-liners, inline.

**Hardware verification before merge**

Since your rig is an inverter and this changes lock behavior on the Powerwall polling path, @jasonacox-sam — please pull this branch and verify against our PW3 (current firmware):

1. **Concurrency**: run the proxy (`ThreadingHTTPServer`) against the PW3 and hammer it with parallel requests across `/vitals`, `/aggregates`, `/pw/status`, and battery-block endpoints — confirm per-method serialization still holds for the single gateway (no duplicate concurrent fetch storms in debug logs, no 429/cooldown triggers).
2. **Soak**: ~30–60 min continuous poll; watch for lock-timeout log lines and stale-cache fallbacks that didn't occur on main.
3. **Byte-compat**: `proxy/regression_test.py` against main on the same hardware — outputs should be identical.
4. **Reconnect under load**: restart/reconnect the proxy while a poll loop is running, to exercise the brief window where a new TEDAPI instance replaces an old one.

Multi-gateway itself can't be exercised on our single PW3 — the unit tests carry that scenario, which is exactly why I want the single-gateway hardware regression to be thorough. Once the two inline items land and Sam's hardware run is clean, I'm happy to merge. This is a solid stepping stone toward multi-gateway support in pypowerwall-server — appreciated.

jasonacox-sam:
Re-checked the branch fresh (`30f3cb9`) — **no changes have landed since the review, so both required one-liners are still open**, and those are the only code items between this PR and the hardware pass:

1. **`pypowerwall/api_lock.py`** (in `acquire_lock_with_backoff`) — drop the misleading annotation so the direct-lock branch reads `lock = lock_holder`; the duck-typing contract is already documented in the docstring above it.
2. **`pypowerwall/tedapi/__init__.py`** (in `_api_lock`) — collapse the lazy-create to the house idiom used in `PyPowerwallCloud._site_api`:
   ```python
   with self._api_locks_guard:
       return self._api_locks.setdefault(name, threading.Lock())
   ```

Non-hardware verification on the current head, so it's on record when the fixes land:

- Full unit suite: **406 passed** locally (py3.12), including all 148 tests in `test_tedapi_cached_fetch.py` — per-instance/per-method lock identity, the lazy-creation race, stale-serve on lock timeout, and all four `acquire_lock_with_backoff` input shapes.

Still pending before merge: the live-hardware pass on our PW3 — parallel polling across `/vitals`, `/aggregates`, `/pw/status`, and battery-block endpoints; a 30–60 min soak watching for lock-timeout/stale-cache fallbacks that don't occur on main; byte-compat against main via `proxy/regression_test.py`; and a proxy restart under load to exercise instance replacement. I'm ready to run all four as soon as the two one-liners land — @jasonacox, requesting your explicit go-ahead for that live-hardware run.

pypowerwall/api_lock.py
Nexarian: Addressed.

jasonacox-sam:
Both required one-liners are in at `a8c3043` — verified against the branch, not the commit message:

1. **Annotation dropped** — `lock = lock_holder` with no `threading.Lock` cast in `acquire_lock_with_backoff`.
2. **`setdefault` idiom** — `_api_lock()` now uses `self._api_locks.setdefault(name, threading.Lock())` under the guard, matching the `PyPowerwallCloud._site_api` pattern.

The Copilot DESIGN.md item is also resolved (`171bce5` — line 123 now documents the per-instance, per-method locking model). Full unit suite on the branch: **659 passed, 10 skipped**.

Code review is now clean from my side. The remaining step is the live-hardware pass against the local TEDAPI gateway — @jasonacox, requesting your approval before I run this on live hardware.

— Sam ⚡

